Issue #2, opened 2023-10-24
feat: Support protocol restrictions on TwingateResource


ekampf:

ekampf:
Fixed by #16

PR #16, opened 2023-10-29
feat: Support protocol restrictions on twingateresource

ekampf:

- Issue: #2 

## Changes

- Modify CRD to support all possible Resource props
- Modify muattions in API later to support these new props

### DIFF
--- a/app/tests/test_crds.py
+++ b/app/tests/test_crds.py
@@ -158,7 +158,7 @@ class TestTwingateResourceCRD:
                 },
             )
 
-    def test_resourceprotocol_validation(self):
+    def test_resourceprotocol_ports_validation(self):
         with pytest.raises(
             ValueError, match="Input should be less than or equal to 65535"
         ):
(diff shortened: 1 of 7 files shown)


app/tests/test_crds.py
```diff
@@ -112,6 +115,88 @@ def test_deserialization(self):
         assert crd.metadata.name == "foo"
         assert crd.metadata.uid == "c560d138-a93a-4463-8b44-d7717851a265"
 
+    def test_is_browser_shortcut_enabled_disallowed_on_wildcard_resource(self):
```
liorr, line 118: What is the expected behavior if ports start > end ?

liorr (approved):
Check my question. I would add another test for that case.  Other than that LGTM.

app/tests/test_crds.py
ekampf: @liorr API call would fail but you're right we can fail early with CRD validation

Issue #5, opened 2023-10-25
chore: Add Issue Templates for Bug/Feature/etc.


ekampf:
Example for reference:
https://github.com/prometheus-operator/prometheus-operator/tree/main/.github/ISSUE_TEMPLATE

ekampf:
Fixed by #17

PR #17, opened 2023-10-29
chore: Add Issue Templates

ekampf:

- Issue: #5 

## Changes

Added issue templates


emrul (approved):
Looks ok to me, if something is wrong we can correct later.

PR #19, opened 2023-10-30
fix: Fix helm chart + add log formatting 

ekampf:
## Changes
- Healthcheck should be on 0.0.0.0 o it doesnt work
- Added log format param to support JSON logs


liorr (approved):
LGTM

PR #23, opened 2023-10-31
test: resourceaccess integration testing

chenbishop:

- Issue:
Adding integration test for resource access flow

## Changes
Adding integration test for resource access flow

## Notes
1. Added environmental variable "TWINGATE_TEST_PRINCIPAL_ID" as requirement for testing as resourceAccess requires the value
2. Added environtal variable "TWINGATE_REMOTE_NETWORK_ID" as requirement for testing as this is required by the operator

### DIFF
--- a/tests_integration/test_main.py
+++ b/tests_integration/test_main.py
@@ -213,6 +213,101 @@ def test_resource_created_before_operator_runs(kopf_settings, unique_resource_na
 
 
 @pytest.mark.integration()
+def test_resource_access_flows(kopf_settings, unique_resource_name):
+    assert "TWINGATE_TEST_PRINCIPAL_ID" in os.environ
+    principal_id = os.environ["TWINGATE_TEST_PRINCIPAL_ID"]
+
+    RESOURCE_OBJ = f"""
+        apiVersion: twingate.com/v1beta
+        kind: TwingateResource
+        metadata:
+          name: {unique_resource_name}
+        spec:
+          name: My K8S Resource
+          address: my.default.cluster.local
+    """
+
+    OBJ = f"""
+        apiVersion: twingate.com/v1beta
+        kind: TwingateResourceAccess
+        metadata:
+          name: {unique_resource_name}
+        spec:
+          resourceRef:
+            name: {unique_resource_name}
+          principalId: {principal_id}
+    """
+
+    # fmt: off
+    with KopfRunner(kopf_runner_args, settings=kopf_settings) as runner:
+        time.sleep(5) # give it some time to react
+        kubectl_create(RESOURCE_OBJ)
+        time.sleep(10)  # give it some time to react
+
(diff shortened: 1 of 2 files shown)


tests_integration/test_main.py
```diff
@@ -542,3 +545,73 @@ def test_protocols_udp_restricted_port_values_start_must_be_lte_end(
...
+        spec:
+          resourceRef:
+            name: {unique_resource_name}
+          principalId: {principal_id}
+    """
+
+    # fmt: off
+    with KopfRunner(kopf_runner_args, settings=kopf_settings) as runner:
+        kubectl_create(RESOURCE_OBJ)
+        time.sleep(10)  # give it some time to react
```
ekampf, line 580: I dont think we need 2 sleeps, we can call `kubectl_create` twice then sleep for `5s`

tests_integration/test_main.py
```diff
@@ -542,3 +545,73 @@ def test_protocols_udp_restricted_port_values_start_must_be_lte_end(
...
+        resource_delete_command = kubectl(f"delete tgr/{unique_resource_name}")
+
+    # fmt: on
+
+    # Ensure that the operator did not die on start, or during the operation.
+    assert runner.exception is None
+    assert runner.exit_code == 0
+
+    logs = load_stdout(runner.stdout)
+    for log in logs:
```
ekampf, line 600: This is a debug leftover - should delete it from `test_resource_flows` too

tests_integration/test_main.py
```diff
@@ -542,3 +545,73 @@ def test_protocols_udp_restricted_port_values_start_must_be_lte_end(
 
         stderr = ex.value.stderr.decode()
         assert "Start port value must be less or equal to end port value" in stderr
+
+
+@pytest.mark.integration()
+def test_resource_access_flows(kopf_settings, unique_resource_name):
+    assert "TWINGATE_PRINCIPAL_ID" in os.environ
```
ekampf, line 552: Lets rename `TWINGATE_PRINCIPAL_ID` to `TWINGATE_TEST_PRINCIPAL_ID` so we know this argument is used for tests

tests_integration/test_main.py
chenbishop: I tried with 5s sleep but the resourceAccess creating was failing (error was the resource not created yet)

tests_integration/test_main.py
ekampf: I added a valid `TWINGATE_TEST_PRINCIPAL_ID` to CI so your PR can pass tests

chenbishop: recommended change applied

tests_integration/test_main.py
chenbishop: recommended change applied

ekampf (approved):

PR #35, opened 2023-11-09
build: Replace pydocstyle with ruff

ekampf:
## Changes

- Removed `pydocstyle`
- Added `pydocstyle` ruff rules (`D`)
- Moved `S101` ignore to be specific to test files


minhtule (approved):
LGTM 👍

PR #36, opened 2023-11-09
fix: chore, build changes should change patch

ekampf:

- Issue:

## Changes
- Added setting that tells `semantic-release` to patch when seeing `chore` or `build` commits:
```
patch_tags = ["fix", "perf", "chore", "build"]
```

Say we release a version `1.0.0` and now dependabot merges a dependency update (`chore` PR) it should be released 
as `1.0.1-dev.<build>` because its newer than `1.0.0`.
Today it gets released as `1.0.0-dev.<build>` which is considered older than `1.0.0`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -162,3 +162,19 @@ prerelease = false
 exclude_commit_patterns = [
     "Bumps*"
 ]
+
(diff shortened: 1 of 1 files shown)


minhtule (approved):
LGTM 👍

PR #38, opened 2023-11-13
test: Simplify integration tests setup

ekampf:
## Changes
- Automatically tag all tests under `tests_integration/` as `integration` so we dont have to manually do it per tests (or group them into classes to do it
- Rename `test_crds` -> `tests_crds_resource` - we'll have a test file per CRD


tests_integration/test_crds_resource.py
```diff
@@ -0,0 +1,345 @@
+import subprocess
```
ekampf, line 1: file just renamed

minhtule (approved):
LGTM 👍

PR #40, opened 2023-11-13
test: Use pytest-factoryboy instead of defining factory fixtures manually

ekampf:
## Changes

Added pytest-factoryboy dependency and integrated it into tests

### Notes
When registering a factory - `ResourceFactory` - the library automatically creates a fixture for it - `resource_factory`
Since this is the same name we were using before - no test changes were required


minhtule (approved):
LGTM 👍

Issue #71, opened 2023-12-15
feat: Allow specifying an apiKey by referencing an existing Secret


ekampf:
**What is missing?**
`values.yaml` requires specifying the `apiKey` directly but for certain customers (dependin on how they manage secrets) it may be better to create the `Secret` object with the apiKey value themselves and point to it.

** Details **
Extend `twingateOperator` section in `values.yaml` to allow either specifying an `apiKey` value or `existingAPIKeySecret`:
Example:
```
twingateOperator:
  network: "acme"
  remoteNetworkId: "123"
  existingAPIKeySecret:
    name: my-secret-object
    key: twingate-api-key

```

PR #72, opened 2023-12-15
feat: Allow specifying API Key via reference to an external Secret

ekampf:

- Issue: #71 

## Changes

Added `existingAPIKeySecret` definition to values.yaml


deploy/twingate-operator/values.schema.json
```diff
@@ -47,9 +47,43 @@
...
+                },
+                {
+                    "required": ["existingAPIKeySecret", "network", "remoteNetworkId"]
+                }
+            ],
             "properties": {
                 "apiKey": { "type": "string" },
+                "existingAPIKeySecret": {
+                    "type": "object",
+                    "title": "An existing secret with the API Key to be used by the operator. If this is specified, the apiKey field will be ignored.",
```
twingate-blee, line 62: I don't think this is correct. When I pass in both `twingateOperator.existingAPIKeySecret.name` and `twingateOperator.apiKey` the apiKey is not ignored. It throws an error

```
helm template --debug -f values.yaml --set twingateOperator.existingAPIKeySecret.name=mysecret --set twingateOperator.existingAPIKeySecret.key=mykey --set twingateOperator.apiKey=test --set twingateOperator.network=test --set twingateOperator.remoteNetworkId=test .

Error: values don't meet the specifications of the schema(s) in the following chart(s):
twingate-operator:
- twingateOperator: Must validate one and only one schema (oneOf)
```

deploy/twingate-operator/values.yaml
```diff
@@ -4,6 +4,9 @@
 
 twingateOperator: {}
```
twingate-blee, line 5: Since apiKey, network, remoteNetworkId are required should we have those uncommented so people don't think they are optional.

Also maybe add a comment to either use apiKey or existingAPIKeySecret.

When not passing either there is an error
```Error: values don't meet the specifications of the schema(s) in the following chart(s):
twingate-operator:
- twingateOperator: Must validate one and only one schema (oneOf)
- twingateOperator: apiKey is required
```
Shoudl that be changed to "apiKey or existingAPIKeySecret1.name is required"

deploy/twingate-operator/values.schema.json
ekampf: You're right. I learned how to use `oneOf` :)
will update

deploy/twingate-operator/values.yaml
ekampf: I dont control that error information it comes from schema validation...

ekampf: Re uncommenting - I want people to explicitly have to uncomment and put values there rather than run with default values which will seem like its working

twingate-blee (approved):
lgtm

PR #93, opened 2024-01-17
fix: Allow setting remote network by name - TwingateSettings failure

ekampf:


minhtule (approved):
LGTM as a temporary solution!

PR #96, opened 2024-01-18
fix: gql.Client can't be singleton as its not thread-safe

ekampf:
## Changes

`gql.Client` is not thread-safe so we cant use it as singleton

https://github.com/graphql-python/gql/issues/314


minhtule (approved):
LGTM 👍

PR #97, opened 2024-01-19
fix: test_remote_network_name_gets_network_id

ekampf:
Should use pytest fixtures and not the `@patch` decorator

### DIFF
--- a/app/tests/test_settings.py
+++ b/app/tests/test_settings.py
@@ -6,6 +6,12 @@ from pydantic import ValidationError
 from app.settings import TwingateOperatorSettings
 
 
+@pytest.fixture()
+def mock_get_remote_network_by_name():
+    with patch("app.api.TwingateAPIClient.get_remote_network_by_name") as m:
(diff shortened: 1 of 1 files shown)


app/tests/test_settings.py
```diff
@@ -6,6 +6,12 @@
 from app.settings import TwingateOperatorSettings
 
 
+@pytest.fixture()
+def mock_get_remote_network_by_name():
+    with patch("app.api.TwingateAPIClient.get_remote_network_by_name") as m:
```
minhtule, line 11: Maybe we want to use [`monkeypatch`](https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html) fixture make this even more pytest-y 😛

minhtule (approved):
LGTM 👍

app/tests/test_settings.py
ekampf: I think it has diff purpose

PR #103, opened 2024-01-23
feat: Add relevant security context to run as non root

### LINKED ISSUE #102: feat: Allow connectors to run under Restricted pod-security policy
**What is missing?**

Add necessary `securityContext` definition to allow connector to run under `restricted` policy enforcement.
Same as this PR in our helm chart - https://github.com/Twingate/helm-charts/pull/41

ekampf:

Resolves #102 

## Changes

Add relevant `securityContext` settings to run connector `0.163.0` as non root and allow running under restricted pod security policy.

### DIFF
--- a/app/handlers/handlers_connectors.py
+++ b/app/handlers/handlers_connectors.py
@@ -54,7 +54,12 @@ def get_connector_pod(
                 "imagePullPolicy": "Always",
                 "name": "connector",
                 "securityContext": {
-                    "allowPrivilegeEscalation": False
+                    "allowPrivilegeEscalation": False,
+                    "capabilities": {
+                        "drop": ["ALL"],
+                    },
+                    "runAsNonRoot": True,
+                    "runAsUser": 65532,
                },
                 **spec.container_extra,
             }


app/handlers/handlers_connectors.py
```diff
@@ -54,7 +54,12 @@ def get_connector_pod(
                 "imagePullPolicy": "Always",
                 "name": "connector",
                 "securityContext": {
-                    "allowPrivilegeEscalation": False
+                    "allowPrivilegeEscalation": False,
+                    "capabilities": {
+                        "drop": ["ALL"],
+                    },
+                    "runAsNonRoot": True,
+                    "runAsUser": 65532,
```
twingate-blee, line 62: We should change this to 1000 since we have an appuser account that the container runs as

```
appuser:x:1000:1000::/home/appuser:/bin/sh
```

ekampf: You are right 🤦‍♂️

ekampf: This image is the **connector** image (distroless, with `nonroot` user whos id is `65532`)

twingate-blee (approved):
lgtm

PR #104, opened 2024-01-23
feat: Add seccompProfile by default

ekampf:
## Changes

Add `seccompProfile` to pod security context by default


twingate-blee (approved):
lgtm

PR #128, opened 2024-02-12
feat: Add support for priorityClassName

### LINKED ISSUE #121: Define `PriorityClass` of operator deployment pods
**What is missing?**

There is no way to define `priorityClassName` of pods created by the Twingate Operator `Deployment`.

**Why do we need it?**

In our cluster, our global default `PriorityClass` is the one we use for our critical applications workloads, forcing us to explicitly define when a workload is more or less important than our services serving production traffic. The Twingate Operator is less important than our application workloads, so we want to give it a lower priority class but the Helm chart doesn't expose a way for use to do that.

**Environment**

    `39d499d684694abd5df762834b7e6a57b81ed9fb`
    

ekampf:

Fixes #121 

## Changes

Adds `priorityClassName` as configurable value to `values.yaml`

### DIFF
--- a/deploy/test/golden/priorityClassName.golden.yaml
+++ b/deploy/test/golden/priorityClassName.golden.yaml
@@ -1,0 +1,145 @@
+---
+# Source: twingate-operator/templates/serviceaccount.yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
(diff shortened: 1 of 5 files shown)


romankor (approved):
GTM

PR #129, opened 2024-02-12
feat: Add podLabels

### LINKED ISSUE #120: Define labels for `Deployment` (and other resources)
**What is missing?**

There is no way to provide a value for `metadata.labels`.

**Why do we need it?**

To integrate with our observability tool (Datadog), we can add labels for auto-discovery of k8s resources. We can forward logs or metrics from the Twingate Operator to our Datadog tenant easily because we cannot add those labels.

**Environment**

    `39d499d684694abd5df762834b7e6a57b81ed9fb`
    

N/A

ekampf:

Fixes #120 

## Changes

Add `podLabels` to `values.yaml` allowing to define additional labels to set on the pod

### DIFF
--- a/deploy/test/golden/podLabels.golden.yaml
+++ b/deploy/test/golden/podLabels.golden.yaml
@@ -1,0 +1,145 @@
+---
+# Source: twingate-operator/templates/serviceaccount.yaml
+apiVersion: v1
+kind: ServiceAccount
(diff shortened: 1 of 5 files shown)


romankor (approved):
gtm

twingate-blee (approved):
lgtm

PR #135, opened 2024-02-14
fix: Disable healthcheck access logs

ekampf:
## Changes

Disable access logs to healthcheck - it just adds a ton of useless logs


romankor (approved):
Gtm

Issue #145, opened 2024-02-15
bug: Resource.is_matching_spec always returns `False`


ekampf:
There's a bug in `protocols` comparison. If no protocol is set API returns the default value 

```
ResourceProtocols(allow_icmp=True, tcp=ResourceProtocol(policy=<ProtocolPolicy.ALLOW_ALL: 'ALLOW_ALL'>, ports=[]), udp=ResourceProtocol(policy=<ProtocolPolicy.ALLOW_ALL: 'ALLOW_ALL'>, ports=[]))
```

But CRD would have `None` as value

PR #146, opened 2024-02-15
fix: Fix resource `protocols` diffing and serialization

ekampf:

- Issue: #145 

## Changes

Fixed `Resource.is_matching_spec` - it was not comparing `protocols` properly so would always return false causing `twingate_resource_sync` to always think there's a drift and issue an update call

### DIFF
--- a/app/api/client_resources.py
+++ b/app/api/client_resources.py
@@ -40,8 +40,8 @@ class ResourceProtocol(BaseModel):
         frozen=True, populate_by_name=True, alias_generator=to_camel
     )
 
-    policy: ProtocolPolicy
-    ports: list[ProtocoRange] | None = None
+    policy: ProtocolPolicy = ProtocolPolicy.ALLOW_ALL
+    ports: list[ProtocoRange] = Field(default_factory=list)
 
 
 class ResourceProtocols(BaseModel):
@@ -49,9 +49,9 @@ class ResourceProtocols(BaseModel):
         frozen=True, populate_by_name=True, alias_generator=to_camel
     )
 
-    allow_icmp: bool | None = None
-    tcp: ResourceProtocol | None = None
-    udp: ResourceProtocol | None = None
+    allow_icmp: bool = True
+    tcp: ResourceProtocol = Field(default_factory=ResourceProtocol)
+    udp: ResourceProtocol = Field(default_factory=ResourceProtocol)
 
 
 class Resource(BaseModel):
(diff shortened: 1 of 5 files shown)


app/api/client_resources.py
```diff
@@ -115,25 +115,27 @@ def is_matching_spec(self, crd: ResourceSpec) -> bool:
...
-            name=self.name,
-            address=self.address.value,
-            alias=self.alias,
-            is_visible=self.is_visible,
-            is_browser_shortcut_enabled=self.is_browser_shortcut_enabled,
-            remote_network_id=self.remote_network.id,
-            security_policy_id=(
-                self.security_policy.id if self.security_policy else None
-            ),
-            protocols=self.protocols,
```
ekampf, line 136: Bug was here - `self.protocols` is an object so we need to serialize it and not send it as is.
Doing `self.protocols.dict()` here would work but a safer approach (what happens if we have another such property in the future?) is to use `self.dict(` and give it the fields we want and it'll take care of it

app/crds.py
```diff
@@ -123,7 +123,11 @@ class ResourceSpec(BaseModel):
     security_policy_id: str | None = None
     is_visible: bool = True
     is_browser_shortcut_enabled: bool = True
-    protocols: ResourceProtocols | None = None
+    protocols: ResourceProtocols = ResourceProtocols(
```
ekampf, line 126: Added default to replicate API behavior - when no protocols limitation is set, reading resource from the API will return a protocols object like this (and not `null`)

app/api/client_resources.py
```diff
@@ -235,7 +244,7 @@ def resource_create(
                 "isBrowserShortcutEnabled": resource.is_browser_shortcut_enabled,
                 "remoteNetworkId": resource.remote_network_id,
                 "securityPolicyId": resource.security_policy_id,
-                "protocols": resource.protocols,
+                "protocols": resource.protocols.dict(by_alias=True),
```
ekampf, line 247: `resource.protocols` is a PyDantic object which is not JSON serializable.
We didnt have a test path here send sends a non-null value before so this never failed

pyproject.toml
```diff
@@ -115,7 +115,7 @@ markers = [
     "integration",
 ]
 
-[tool.ruff]
```
ekampf, line 118: Fix ruff warnings re configuration location change

liorr (approved):
LGTM

Issue #161, opened 2024-02-24
Twingate Connector doesn't handle pod deletion on "resume"


ekampf:
**What happened?**

In various cases (like cluster upgrade) pod can go down while operator is down and wont get the event.
When operator is back, we dont check for pod existence on resume.
Operator will think connector is fine while in fact pod is missing

PR #162, opened 2024-02-24
fix: twingate connector doesnt handle pod deletion on resume

ekampf:

- Issue: #161 

## Changes

On TwingateConnector resume event check if pod exists


liorr (approved):
LGTM

PR #175, opened 2024-02-29
fix: Limit twingate_resource_access_update scope to spec

### LINKED ISSUE #174: fix: twingate_resource_access_update listens to too many updates bug
`twingate_resource_access_update` should only care about `spec` updates.

ekampf:

Fixes #174 

## Changes

Limit the update scope of `twingate_resource_access_update` to just the `spec` field

### DIFF
--- a/app/handlers/handlers_resource_access.py
+++ b/app/handlers/handlers_resource_access.py
@@ -47,10 +47,10 @@ def twingate_resource_access_create(body, spec, memo, logger, patch, **kwargs):
 
 
 @kopf.on.update("twingateresourceaccess", field="spec")
-def twingate_resource_access_update(spec, new, diff, status, memo, logger, **kwargs):
+def twingate_resource_access_update(spec, diff, status, memo, logger, **kwargs):
     logger.info(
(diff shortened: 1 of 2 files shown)


app/handlers/handlers_resource_access.py
```diff
@@ -46,8 +46,8 @@ def twingate_resource_access_create(body, spec, memo, logger, patch, **kwargs):
         return fail(error=mex.error)
 
 
-@kopf.on.update("twingateresourceaccess")
-def twingate_resource_access_update(new, diff, status, memo, logger, **kwargs):
+@kopf.on.update("twingateresourceaccess", field="spec")
+def twingate_resource_access_update(spec, new, diff, status, memo, logger, **kwargs):
     logger.info(
         "Got TwingateResourceAccess update request: %s. Diff: %s. Status: %s.",
         new,
```
minhtule, line 53: Maybe we could just log the spec instead of new and remove that argument?

minhtule (approved):
LGTM 👍

PR #176, opened 2024-02-29
fix: Reduce update scope for `twingate_resource_update`

### LINKED ISSUE #173: fix: twingate_resource_update listens to too many updates
`twingate_resource_update` is too broad and should optimize it to happen less:
1. Only when spec happens
2. Try to figure out a way not to trigger an ipdate when `id` changes from `None` to a value following create handler

ekampf:

Fixes #173 

## Changes

`twingate_resource_update` should only care when `spec` updates

### DIFF
--- a/app/handlers/handlers_resource.py
+++ b/app/handlers/handlers_resource.py
@@ -4,7 +4,7 @@ import kopf
 
 from app.api import TwingateAPIClient
 from app.crds import ResourceSpec
-from app.handlers.base import success
+from app.handlers.base import fail, success
 
 
 @kopf.on.create("twingateresource")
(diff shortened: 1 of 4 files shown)


app/handlers/handlers_resource.py
```diff
@@ -21,16 +21,20 @@ def twingate_resource_create(body, spec, memo, logger, patch, **kwargs):
     )
 
 
-@kopf.on.update("twingateresource")
-def twingate_resource_update(old, new, diff, status, memo, logger, **kwargs):
+@kopf.on.update("twingateresource", field="spec")
+def twingate_resource_update(spec, new, diff, status, memo, logger, **kwargs):
     logger.info(
         "Got TwingateResource update request: %s. Diff: %s. Status: %s.",
```
minhtule, line 27: Maybe we could just log the `spec` instead of `new` and remove that argument?

app/handlers/handlers_resource.py
```diff
@@ -21,16 +21,20 @@ def twingate_resource_create(body, spec, memo, logger, patch, **kwargs):
...
         new,
         diff,
         status,
     )
 
-    crd = ResourceSpec(**new["spec"])
+    crd = ResourceSpec(**spec)
+
+    if len(diff) == 1 and diff[0][0] == "add" and diff[0][1] == ("id",):
```
minhtule, line 35: nit: I wonder if there is a nicer way to do this because the indexes are quite abstract. Maybe
```python
if diff == (("add", ("id",), None, ANY),):
```
? 🤔

app/handlers/tests/test_handlers_resource.py
```diff
@@ -100,6 +91,42 @@ def test_update(self, mock_api_client):
...
         assert patch_mock.spec == {}
 
+    def test_update_caused_by_create_does_nothing(self, mock_api_client):
+        rid = "UmVzb3VyY2U6OTMxODE3"
+        spec = new = {
+            "id": rid,
+            "address": "my.default.cluster.local",
+            "name": "new-name",
+        }
+        diff = (("add", ("id",), rid),)
```
minhtule, line 101: nit: I think even for `add` action, there is still the old value of `None` ([doc](https://kopf.readthedocs.io/en/stable/walkthrough/diffs/#diffs)) so this should be
```python
diff = (("add", ("id",), None, rid),)
```

minhtule (approved):
LGTM 👍

app/handlers/handlers_resource.py
ekampf: @minhtule I thought about it but `ANY` is in `unittest` which I wouldnt want to import on non testing code

PR #180, opened 2024-03-01
feat: Improved TwingateConnector reconciliation

ekampf:
## Changes

Replaced the complex mechanism of events we had to control TwingateConnector's pod lifecycle with a simple timer that runs every 5s and makes sure pod is alive and up to date


app/handlers/handlers_connectors.py
```diff
@@ -292,44 +158,49 @@ def twingate_connector_delete(spec, meta, status, namespace, memo, logger, **kwa
...
-            owner_version,
-            namespace,
-            "twingateconnectors",
-            owner["name"],
-            {"metadata": {"labels": {LABEL_CONNECTOR_POD_DELETED: "true"}}},
+    if crd.spec.image or not k8s_pod:
+        image = crd.spec.get_image()
+    elif crd.spec.image_policy:
+        now = pendulum.now("UTC").start_of("minute")
+        next_check = pendulum.parse(
```
minhtule, line 187: nit: I would prefix this with `_at` to make it obvious that this is a point in time and not a duration which I initially thought 😳

minhtule (approved):
LGTM 👍

PR #184, opened 2024-03-05
fix: Connector pod updates "Forbidden" error

ekampf:
## Changes

If we change `TwingateConnector` properties (`logLevel` for example) we'll get the following error:
```
Forbidden: pod updates may not change fields other than `spec.containers[*].image
```

So this PR makes the following changes:
1. Reconciler only checks the container image if pod exists
2. When `TwingateConnector` updates we delete the pod and reconcile it


minhtule (approved):
LGTM 👍

PR #188, opened 2024-03-07
chore: Set poetry to package-mode = false

ekampf:

Fixes the following warning when installing dependencies:
```
Warning: The current project could not be installed: No file/folder found for package twingate-operator
If you do not want to install the current project use --no-root.
If you want to use Poetry only for dependency management but not for packaging, you can disable package mode by setting package-mode = false in your pyproject.toml file.
In a future version of Poetry this warning will become an error!
```


minhtule (approved):
LGTM 👍

PR #189, opened 2024-03-07
fix: Fix k8s_force_delete_pod finalizers removal

ekampf:
## Changes

`k8s_force_delete_pod` was not deleting pod finalizers.
`[]` makes it merge the value with existing value instead of deleting it.


minhtule (approved):
LGTM 👍

PR #190, opened 2024-03-07
chore: Integration test for old pod migration

ekampf:
## Changes

Tests the fix of #189 by adding `test_connector_flows_pod_migration_from_older_pod_with_finalizers`.
We create a connector, patch the pod with a finalizer and wrong version annotation and make sure the reconciler is able to delete and recreate it.


minhtule (approved):
LGTM 👍

Issue #192, opened 2024-03-11
Enable to block all ports


arthur-leclerc:
**What happened?**

I'm unable to block all ports on a `TwingateResource` when using:

```yaml
  protocols:
    udp:
      policy: RESTRICTED
      ports: []
```

**Did you expect to see something different?**

I'm pretty sure that's the way to do it though. When I look at the call from the website, the GraphQL call looks like this:

<img width="214" alt="Capture d’écran 2024-03-12 à 00 00 36" src="https://github.com/Twingate/kubernetes-operator/assets/43139375/6d9c537c-1ad2-4bbb-ac68-79f042c90254">

**How to reproduce it (as minimally and precisely as possible)**:

**Environment**

* Twingate Kubernetes Operator version:

    `twingate/kubernetes-operator:0.6.1`
    

* Kubernetes version information:

```
Client Version: v1.29.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.29.1-eks-508b6b3
```

* Kubernetes cluster kind:

    AWS EKS

* Manifests:

```yaml
apiVersion: twingate.com/v1beta
kind: TwingateResource
metadata:
  name: foo
spec:
  name: Foo
  address: foo.default.cluster.local
  protocols:
    allowIcmp: true
    tcp:
      policy: RESTRICTED
      ports:
        - end: 8080
          start: 8080
    udp:
      policy: RESTRICTED
      ports: []
```

* Twingate Kubernetes Operator Logs:

```
Invalid value: "object": Can't specify port ranges for ALLOW_ALL policy, and must specify port ranges for RESTRICTED policy
```

**Anything else we need to know?**:

ekampf:
Hey @arthur-leclerc confirming this is a bug and will issue a fix ASAP

arthur-leclerc:
Works as expected now, thanks.

PR #193, opened 2024-03-11
fix: Resource restricted policy should allow empty ports

ekampf:

- Issue: #192 

## Changes

- Fixed minimum port value - should be 1 not zero
- Allow `RESTRICTED` ports to be empty list which means "block all"

Also manually tested adding this resource:
```
kind: TwingateResource
metadata:
  name: foo
spec:
  name: Foo
  address: foo.default.cluster.local
  protocols:
    allowIcmp: true
    tcp:
      policy: RESTRICTED
      ports:
        - end: 8080
          start: 8080
    udp:
      policy: RESTRICTED
      ports: []
```
result:
Screenshot 2024-03-11 at 4 47 03 PM

### DIFF
--- a/app/crds.py
+++ b/app/crds.py
@@ -68,8 +68,8 @@ class ProtocolRange(BaseModel):
         frozen=True, populate_by_name=True, alias_generator=to_camel
     )
 
-    start: int = Field(ge=0, le=65535)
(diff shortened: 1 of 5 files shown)


minhtule (approved):
LGMT 👍

PR #223, opened 2024-03-26
feat: Make TwingateConnector's containerExtra and podExtra mutable

### LINKED ISSUE #215: Make connector `containerExtra` mutable
**What is missing?**

Once the `containerExtra` object is set, it can not be altered. I don't have a good reason I see to make it immutable and it makes updating it painful since it requires a full resource delete and create again.

**Why do we need it?**

Lacking any example documentation on the field, I resorted to trial and error to get the desired result. The immutable aspect makes that process cumbersome and frankly annoying.

**Environment**

    `0.6.2`
    

I'm resorting to this because I want to enable the log analytics. It would be nice for this to be a first class config option for the resource.

ekampf:

Fixes #215 

## Changes

- Make `containerExtra` and `podExtra` mutable

The refactor in `TwingateConnector` reconciliation (https://github.com/Twingate/kubernetes-operator/pull/180) made this possible out of the box because we delete the pod and recreate it anyway on any update

### DIFF
--- a/deploy/twingate-operator/crds/twingate.com.twingateconnector.yaml
+++ b/deploy/twingate-operator/crds/twingate.com.twingateconnector.yaml
@@ -19,10 +19,6 @@ spec:
               x-kubernetes-validations:
                 - rule: (!has(oldSelf.id) || self.id == oldSelf.id)
                   message: "id is immutable once set"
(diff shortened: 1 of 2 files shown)


minhtule (approved):
LGTM 👍

PR #225, opened 2024-03-26
feat: TwingateConnector - allow defining extra pod annotations with `podAnnotations`

### LINKED ISSUE #217: Connector pod annotation configuration
**What is missing?**

Ability to set pod annotations on the create pod for a connetor.

**Why do we need it?**

For Datadog log monitoring, we need to configure the agent using annotations on the pod.

**Environment**

    `0.6.2`
    

ekampf:

Fixes #217 

## Changes

- Add `podAnnotations` to `TwingateConnector` CRD
- propagate annotations to created pod

### DIFF
--- a/app/handlers/handlers_connectors.py
+++ b/app/handlers/handlers_connectors.py
@@ -74,8 +74,9 @@ def get_connector_pod(
         ],
         **spec.pod_extra,
     }
+    pod_annotations = spec.pod_annotations.copy().update({ANNOTATION_POD_SPEC_VERSION: ANNOTATION_POD_SPEC_VERSION_VALUE})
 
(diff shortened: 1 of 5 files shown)


app/handlers/handlers_connectors.py
```diff
@@ -74,8 +74,9 @@ def get_connector_pod(
         ],
         **spec.pod_extra,
     }
+    pod_annotations = spec.pod_annotations.copy().update({ANNOTATION_POD_SPEC_VERSION: ANNOTATION_POD_SPEC_VERSION_VALUE})
```
minhtule, line 77: I think `dict.update()` would return `None` so we probably want something like
```python
pod_annotations = spec.pod_annotations | {ANNOTATION_POD_SPEC_VERSION: ANNOTATION_POD_SPEC_VERSION_VALUE}
```

minhtule (approved):
Fix & merge!

PR #234, opened 2024-04-02
feat: Add TwingateConnector sidecarContainers prop

### LINKED ISSUE #232: Support providing additional container definitions to support sidecar deployments
**What is missing?**
Running "sidecar" containers is a common pattern in Kubernetes.  So much so that Kubernetes [1.29 even supports](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) them as a "first class citizen".

There is currently [no way in the connector handler](https://github.com/Twingate/kubernetes-operator/blob/main/app/handlers/handlers_connectors.py#L73-L74) to include specifications for additional containers that should run as part of the pod.

**Why do we need it?**
The need may arise for folks to include additional containers to run in the pod the operator creates for a _TwingateConnector_ for things such as (but not limited to):
* Log Shipping
* Monitoring
* Proxy control (i.e. Envoy)

A more concrete use-case would be if you require a log-shipping container to run inside of the pod the connector runs in to ship the logs from the connector to a 3rd party.  This could take the form of a fluentd sidecar, a datadog-agent sidecar, etc.

We'd need the flexibility to essentially submit a full container specification to be generic enough for many use-cases.

**Environment**

    [`v0.7.0`](https://github.com/Twingate/kubernetes-operator/releases/tag/v0.7.0)
    

I don't believe we should rely on the _native_ kubernetes sidecar implementation, because that would require at minimum v. 1.29 of kubernetes, and the feature itself is still in _beta_ as of that version.  Most folks are used to sidecar patterns without the native kubernetes support.

ekampf:

Resolves #232

## Changes

Added `sidecarContainers` to allow injecting sidevcars to connector pods

## Notes

Example on prometheus operator: 
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L3853

### DIFF
--- a/app/crds.py
+++ b/app/crds.py
@@ -306,6 +306,7 @@ class ConnectorSpec(BaseModel):
     container_extra: dict[str, Any] = {}
     pod_extra: dict[str, Any] = {}
     pod_annotations: dict[str, Any] = {}
+    sidecar_containers: list[dict[str, Any]] = []
 
     remote_network_id: str = Field(
         default_factory=lambda: get_settings().remote_network_id
--- a/app/handlers/handlers_connectors.py
+++ b/app/handlers/handlers_connectors.py
@@ -65,7 +65,8 @@ def get_connector_pod(
                     "runAsUser": 65532,
                },
                 **container_extra,
-            }
+            },
+            *spec.sidecar_containers,
         ],
         **spec.pod_extra,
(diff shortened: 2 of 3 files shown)


minhtule (approved):
LGTM 👍

PR #236, opened 2024-04-04
feat: Add ability to customize kopf logging level

ekampf:
## Changes

- Added `twingateOperator.logVerbosity` to helm chart's `values.yaml`

### DIFF
--- a/deploy/twingate-operator/templates/deployment.yaml
+++ b/deploy/twingate-operator/templates/deployment.yaml
@@ -47,9 +47,11 @@ spec:
           - ./main.py
           - "-A"
           - "--standalone"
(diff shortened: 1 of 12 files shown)


deploy/twingate-operator/templates/deployment.yaml
```diff
@@ -47,7 +47,7 @@ spec:
           - ./main.py
           - "-A"
           - "--standalone"
-          - "--verbose"
+          - "--{{- .Values.twingateOperator.logVerbosity | default "verbose" -}}"
```
rlafferty, line 50: I believe this will work _except for_ if you want `INFO` log level, which the assumption in this code is that no argument will be supplied: https://github.com/zalando-incubator/kopf/blob/master/kopf/engines/logging.py#L142

PR #243, opened 2024-04-08
feat: Don’t query for principal external ref once we have the id

### LINKED ISSUE #239: feat: Optimize TwingateResourceAccess reconciliation and handle name changes
**What is missing?**
Problems:
- `TwingateResourceAccess` reconciliation is very chatty when `principalExternalRef` is used - searching for the name and reapplying it on every run
- It also doesnt handle well a case where group name changes (fails) or replaced (now both groups will have access)

Proposal:

- When you specify `principalExternalRef` and `princiaplId` is null we'll search for that name, get the id and modify the `TwingateResourceAccess.spec.princiaplId` to that value.
- going forward as long as principalId has value `principalExternalRef` is ignored
- (?) `principalExternalRef` is not used for reconciliation and might hold an old outdated name which might be confusing when debugging. How can we solve this?  one way would be to reconcile the nam from `principalId` referenced entity to `principalExternalRef`

**Why do we need it?**

ekampf:

Resolves #239 

## Changes

- create\update\sync handler were all basically doing the same thing - removed redundant code and just use a single function
- When `principalExternalRef` is used - cache the resulting `principalId` on creation and use that value going forward

### DIFF
--- a/app/handlers/handlers_resource_access.py
+++ b/app/handlers/handlers_resource_access.py
@@ -51,14 +51,6 @@ def check_status_created(status: dict | None) -> dict | None:
     return None
 
 
-@kopf.on.create("twingateresourceaccess")
-@kopf.on.update("twingateresourceaccess", field="spec")
-@kopf.timer(
-    "twingateresourceaccess",
-    interval=timedelta(seconds=10).seconds,
-    initial_delay=60,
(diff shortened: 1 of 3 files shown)


minhtule (approved):
Great refactor! 👍

PR #244, opened 2024-04-10
feat: Allow exposing Service objects with annotations

### LINKED ISSUE #21: Feature Proposal: Add resources by annotating Service objects
To make services accessible on K8S a `Service` object is used.
Rather than make user manually create an aditional `TwingateResource` to expose that service we can do that automatically.

## Implementation details

- To expose a `Service` as a Twingate Resource, annotate it with `twingate.com/expose` value `"true"` (under its `metadata.annotations`)
- Operator will listen on annotated services and create a child `TwingateResource` object with address `<service name>.<namespace>.svc.cluster.local`

### Questions
- What about other `TwingateResource` properties like `isvisible`, `isBrowserShortcutEnabled` and `securityPolicyId` ?  Maybe the  `twingate.com/expose` should actually be an object...
- What if you have the same service on multiple clusters? (prod\non prod) they will all get the same resource address...

ekampf:

Resolves #21 

## Changes

Added `handlers_service.py` to listen to K8s `Service` objects with annotations and create the corresponding `TwingateResource`

### DIFF
--- a/app/handlers/handlers_services.py
+++ b/app/handlers/handlers_services.py
@@ -1,0 +1,50 @@
+import kopf
+import kubernetes
+
+from app.handlers import success
+
+
+@kopf.on.resume("services", annotations={"twingate.com/expose": "true"})
+@kopf.on.create("service", annotations={"twingate.com/expose": "true"})
+def twingate_service_create(body, spec, namespace, meta, logger, **_):
+    logger.info("twingate_service_create: %s", spec)
+    service_name = body.meta.name
+
+    resource_subobject = {
+        "apiVersion": "twingate.com/v1beta",
+        "kind": "TwingateResource",
+        "metadata": {
+            "name": f"{service_name}-resource",
+        },
+        "spec": {
+            "name": f"{service_name}-resource",
+            "address": f"{service_name}.{namespace}.svc.cluster.local",
+        },
+    }
+    kopf.adopt(resource_subobject)
+
+    if alias := meta.annotations.get("twingate.com/expose-alias"):
+        resource_subobject["spec"]["alias"] = alias
+
+    if service_ports := spec.get("ports", []):
+        protocols = {
+            "tcp": {"policy": "RESTRICTED", "ports": []},
+            "udp": {"policy": "RESTRICTED", "ports": []},
+        }
+        for port_obj in service_ports:
+            port = port_obj["port"]
+            if port_obj["protocol"] == "TCP":
+                protocols["tcp"]["ports"].append({"start": port, "end": port})
+            elif port_obj["protocol"] == "UDP":
+                protocols["udp"]["ports"].append({"start": port, "end": port})
+
+        resource_subobject["spec"]["protocols"] = protocols
+
+    kapi = kubernetes.client.CustomObjectsApi()
+    api_response = kapi.create_namespaced_custom_object(
+        "twingate.com", "v1beta", namespace, "twingateresources", resource_subobject
+    )
+    logger.info("create_namespaced_custom_object response: %s", api_response)
+    return success(
+        message=f"Created TwingateResource {resource_subobject['spec']['name']}"
+    )
--- a/app/handlers/__init__.py
+++ b/app/handlers/__init__.py
@@ -2,3 +2,4 @@
 from .handlers_connectors import *
 from .handlers_resource import *
 from .handlers_resource_access import *
+from .handlers_services import *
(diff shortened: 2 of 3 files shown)


app/handlers/handlers_services.py
```diff
@@ -0,0 +1,94 @@
...
+
+@kopf.on.resume("services", annotations={"twingate.com/expose": "true"})
+@kopf.on.create("service", annotations={"twingate.com/expose": "true"})
+@kopf.on.update("service", annotations={"twingate.com/expose": "true"})
+def twingate_service_create(body, spec, namespace, meta, logger, **_):
+    logger.info("twingate_service_create: %s", spec)
+    service_name = body.meta.name
+    resource_object_name = f"{service_name}-resource"
+
+    resource_subobject = service_to_twingate_resource(body, namespace)
```
minhtule, line 69: The `resource_object_name` is also created in `service_to_twingate_resource()` which might lead to inconsistency. I would either pass this `resource_object_name` to `service_to_twingate_resource()`, or having `service_to_twingate_resource()` method return this value.

app/handlers/handlers_services.py
```diff
@@ -0,0 +1,94 @@
...
+                protocols["tcp"]["ports"].append({"start": port, "end": port})
+            elif port_obj["protocol"] == "UDP":
+                protocols["udp"]["ports"].append({"start": port, "end": port})
+
+        result["spec"]["protocols"] = protocols
+
+    return result
+
+
+@kopf.on.resume("services", annotations={"twingate.com/expose": "true"})
```
minhtule, line 61: nit: we could remove the trailing `s` just for consistency

app/handlers/handlers_services.py
```diff
@@ -0,0 +1,94 @@
...
+    resource_object_name = f"{service_name}-resource"
+
+    resource_subobject = service_to_twingate_resource(body, namespace)
+    kopf.adopt(resource_subobject)
+
+    kapi = kubernetes.client.CustomObjectsApi()
+    existing_resource_object = k8s_get_twingate_resource(
+        namespace, resource_object_name, kapi
+    )
+    if existing_resource_object:
```
minhtule, line 76: nit: no walrus operator? 😛

app/handlers/tests/test_handlers_services.py
```diff
@@ -0,0 +1,184 @@
...
+                "udp": {
+                    "policy": "RESTRICTED",
+                    "ports": [{"start": 22, "end": 22}],
+                },
+            },
+        },
+    }
+
+
+def test_service_to_twingate_resource_without_alias(example_service_body):
```
minhtule, line 85: nit: Would be nice to create test case for `service_to_twingate_resource` function to group this and the test above.

app/handlers/tests/test_handlers_services.py
```diff
@@ -0,0 +1,184 @@
...
+    )
+    with pytest.raises(kubernetes.client.exceptions.ApiException):
+        k8s_get_twingate_resource("default", "test")
+
+
+class TestTwingateServiceCreate:
+    def test_create_service(
+        self, example_service_body, kopf_handler_runner, k8s_customobjects_client_mock
+    ):
+        service_body = example_service_body
```
minhtule, line 137: nit: we probably could name the factory `service_body` to avoid this renaming here

app/handlers/tests/test_handlers_services.py
```diff
@@ -0,0 +1,184 @@
...
+):
+    k8s_customobjects_client_mock.get_namespaced_custom_object.side_effect = (
+        kubernetes.client.exceptions.ApiException(status=500)
+    )
+    with pytest.raises(kubernetes.client.exceptions.ApiException):
+        k8s_get_twingate_resource("default", "test")
+
+
+class TestTwingateServiceCreate:
+    def test_create_service(
```
minhtule, line 134: I think this name should be `test_create_twingate_resource` because we're not creating k8s service

app/handlers/tests/test_handlers_services.py
```diff
@@ -0,0 +1,184 @@
...
+        k8s_customobjects_client_mock.patch_namespaced_custom_object.assert_not_called()
+        k8s_customobjects_client_mock.create_namespaced_custom_object.assert_called_once_with(
+            "twingate.com",
+            "v1beta",
+            "default",
+            "twingateresources",
+            service_to_twingate_resource(service_body, "default"),
+        )
+
+    def test_update_service(
```
minhtule, line 158: Should be something like `test_update_existing_twingate_resource`

minhtule (approved):
Fix & merge!

PR #270, opened 2024-05-03
fix: using the operator with annotations have some problems

### LINKED ISSUE #269: Using the Operator with annotations have some problems
Hi
I use the newest operator and try to create resources with annotations.
At first, I added the: `twingate.com/resource: 'true'` annotation and got:

> cannot patch resource "services/status" in API group "" in the namespace "namespace"', 'reason': 'Forbidden'

So I added `services/status` manually to the clusterRole.

Then I got:

>  cannot create resource \"twingateresources\" in API group \"twingate.com\" in the namespace \"namespace\"","reason":"Forbidden"  

So I added 'create' manually to the clusterRole.

Then the resource was created.
Alias works fine.

When I tried to add isVisible annotation, I got:

> spec.isVisible: Invalid value: \\\\\\"string\\\\\\": spec.isVisible in body must be of type boolean

but k8s doesn't let you enter a non-string value in annotations. Both "False" and "false" didn't work.

Please kindly fix it.
Thanks!

ekampf:

Fixes #269 

## Changes

- Operator annotation values need to be converted to the right type
- Added missing permissions to cluster role

### DIFF
--- a/app/handlers/handlers_services.py
+++ b/app/handlers/handlers_services.py
@@ -1,5 +1,6 @@
 import kopf
 import kubernetes
+from attr.converters import to_bool
 
 from app.handlers import success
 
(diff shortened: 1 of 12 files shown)


minhtule (approved):
LGTM :thum

PR #289, opened 2024-05-16
fix: Remove TwingateResourceAccess ownerReference

### LINKED ISSUE #287: ownerReferences of TwingateResourceAccess prevents ArgoCD from deleting it
Using v0.11.2. We are using ArgoCD to create both TwingateResource and TwingateResourceAccess.

When I delete a TwingateResourceAccess from the source code, that what I see in Argo:
twingatepic
The resource on the right (baseline-analyzer-developers-access) was before the deletion like other resources (baseline-analyzer-admins-access for instance).
Once we delete it, ArgoCD refuse to delete it as it has the ownerReferences field, that connects it to the resource.
Once we delete the ownerReferences field, the resource is deleted.

I came across this: https://github.com/argoproj/argo-cd/issues/4764
It may help...
Thanks!

ekampf:

Fix #287 

## Changes

- Remove ownerReference for `TwingateResourceAccess` 

If a resource that the access object is referencing gets removed the access object shouldnt get deleted as well but remain there and log errors on reconciliation

### DIFF
--- a/app/handlers/handlers_resource_access.py
+++ b/app/handlers/handlers_resource_access.py
@@ -81,9 +81,6 @@ def twingate_resource_access_change(body, spec, memo, logger, patch, status, **k
             reason="Success",
             message=f"Added access to {resource_crd.spec.id}<>{principal_id}",
         )
-        patch.metadata["ownerReferences"] = [
-            resource_crd.metadata.owner_reference_object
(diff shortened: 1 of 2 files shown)


minhtule (approved):
LGTM 👍

Issue #364, opened 2024-07-23
Support group creation and reference


ekampf:
**What is missing?**
Often when creating Twimgate resources and we also want to give access to it to a group of users.
Often, these groups will be created somewhere else - synced from the Identity Provider (Okta etc.).
But there's also a use case we often see in the TF world of creating a manual group for this purpose and we should support it in the k8s operator the same way we support it in TF.

- Allow creating groups via a `TwingateGroup` CRD
  - Support referencing users by `id`
  - Support referencing users by `email`
- Allow referencing a `TwingateGroup` instance on a `TwingateResourceAccess` instance

Example:

```
apiVersion: twingate.com/v1beta
kind: TwingateGroup
metadata:
  name: my-group
spec:
  name: My K8S Group
  members: ["VXNlcjoxMjMK", "foo@bar.com"]
---
apiVersion: twingate.com/v1beta
kind: TwingateResourceAccess
metadata:
  name: my-twingate-access
spec:
  resourceRef:
    name: my-twingate-resource
    namespace: default
  groupRef:
    name: my-group
    namespace: default
```

Open questions: 
- Should `TwingateGroup` enforce group membership members to be only the members described?  at at-least the members described? (`authoritative: true\false`)

ErezWeiss:
We are looking forward to it!

PR #367, opened 2024-07-24
feat: TwingateGroup to create\update groups via operator

ekampf:

- Issue: #364 

## Changes

- Added `TwingateGroup` CRD and functionality


app/crds.py
```diff
@@ -339,3 +341,62 @@ class TwingateConnectorCRD(BaseK8sModel):
...
+    security_policy_id: str | None = None
+    members: set[str] = set()
+
+    @cached_property
+    def members_hash(self) -> str:
+        members_str = ",".join(sorted(self.members))
+        return hashlib.md5(members_str.encode()).hexdigest()  # noqa: S324 # nosec
+
+    @cached_property
+    def members_email(self) -> set[str]:
```
minhtule, line 364: nit: should have trailing `s` like `members_ids`?

app/handlers/handlers_groups.py
```diff
@@ -0,0 +1,85 @@
...
+
+    if diff := kwargs.get("diff"):
+        logger.info("Diff: %s", diff)
+        # If ID changed from `None to value we just created it no need to update
+        # diff is `(('add', ('spec', 'id'), None, 'R3JvdXA6MjAxNjc4OQ=='),)`
+        if (
+            len(diff) == 1
+            and diff[0][0] == "add"
+            and diff[0][1] == ("spec", "id")
+            and diff[0][2] is None
```
minhtule, line 27: nit: maybe `diff[0][:3] == ("add", ("spec", "id"), None)` like in the [resource update handler](https://github.com/Twingate/kubernetes-operator/blob/d143995d46f1a381d0571dfb05df9843d919abcd/app/handlers/handlers_resource.py#L51)

app/handlers/handlers_groups.py
```diff
@@ -0,0 +1,85 @@
...
+        ):
+            return success()
```
minhtule, line 29: Should we provide some information like
```
return success(
    twingate_id=group.id,
    created_at=group.created_at.isoformat(),
    updated_at=group.updated_at.isoformat(),
)
```
like in other handler functions?

app/api/client_users.py
```diff
@@ -0,0 +1,42 @@
...
+                            edges { node { id, email } }
+                        }
+                    }
+                    """),
+                    variable_values={"emails": emails, "after": after},
+                )
+                users = page["users"]
+                page_info = users["pageInfo"]
+                for edge in users["edges"]:
+                    result[edge["node"]["email"]] = edge["node"]["id"]
```
minhtule, line 32: Some IdPs like AuzreAD doesn't enforce email uniqueness so this might result in matching on the first matched user.

app/crds.py
```diff
@@ -339,3 +341,62 @@ class TwingateConnectorCRD(BaseK8sModel):
...
+    email: str | None = None
+
+
+class GroupStatus(BaseModel):
+    model_config = ConfigDict(
+        frozen=True, populate_by_name=True, alias_generator=to_camel, extra="allow"
+    )
+
+    user_ids: list[GroupStatusUserID] = []
+    user_ids_hash: str | None = None
```
minhtule, line 387: I would rename this to `members_hash` because with this name, I would expect this value to be the has of the `user_ids` list above

app/handlers/handlers_groups.py
```diff
@@ -0,0 +1,85 @@
...
+        logger.info("User IDs cache is invalid, fetching from Twingate")
+        kopf.info(body, reason="User IDs cache is invalid, fetching from Twingate")
+
+        email_to_id = client.get_ids_for_emails(crd.spec.members_email)
+        user_ids = [
+            GroupStatusUserID(id=id, email=email) for email, id in email_to_id.items()
+        ]
+        user_ids.extend([GroupStatusUserID(id=uid) for uid in crd.spec.members_ids])
+        patch.status["user_ids"] = [ui.model_dump() for ui in user_ids]
+        patch.status["user_ids_hash"] = crd.spec.members_hash
```
minhtule, line 43: I think there is an edge case in caching the email lookup. Let's say the user's email is `foo@gmail.com` and the group k8s spec is 
```yaml
spec:
  name: Example Group
  members:
    - "foo@gmail.com"
```
If that user's email is changed to `bar@gmail.com`, the user still belongs to this `Example Group`. This doesn't happen very often though. Maybe we should have a cache expiration like 1 day or 1 week? 🤔

deploy/twingate-operator/crds/twingate.com.twingategroup.yaml
```diff
@@ -0,0 +1,47 @@
...
+                name:
+                  type: string
+                  description: "Name of the group."
+                securityPolicyId:
+                  type: string
+                  nullable: true
+                  pattern: "^[-A-Za-z0-9+/]*={0,3}$"
+                members:
+                    type: array
+                    x-kubernetes-list-type: set
```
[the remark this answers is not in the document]
ekampf: it just makes k8s validate no duplicate and give nicer error message if there are

PR #397, opened 2024-09-05
feat: TwingateGroup object - support creating groups via k8s object

ekampf:

- Issue: #364 

## Changes

- Add `TwingateGroup` CRD that supports creating\deleteing a group and updating its name


app/api/tests/test_client_groups.py
```diff
@@ -43,3 +47,210 @@ def test_get_group_id_not_found_returns_none(
...
+                    {"variables": {"name": name, "userIds": []}}, strict_match=False
+                )
+            ],
+        )
+        result = api_client.group_create(GroupSpec(name=name))
+        assert result == "test-id"
+
+    def test_group_create_failure(self, test_url, api_client, mocked_responses):
+        name = "my group name"
+        success_response = json.dumps(
```
minhtule, line 79: nit: should be `failed_response`?

app/api/client_groups.py
```diff
@@ -31,3 +88,53 @@ def get_group_id(self: TwingateClientProtocol, group_name: str) -> str | None:
...
+            result = self.execute_mutation(
+                "groupDelete",
+                MUT_DELETE_GROUP,
+                variable_values={"id": group_id},
+            )
+
+            return bool(result["ok"])
+        except GraphQLMutationError as gql_err:
+            if "does not exist" in gql_err.error:
+                return True
```
minhtule, line 136: I think we don't need this logic anymore. We update the Public API to consistently return success (with `ok=True`) if an entity doesn't exist for all kinds of entity in https://github.com/Twingate/controller/pull/8735

app/handlers/handlers_groups.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if diff := kwargs.get("diff"):
+        logger.info("Diff: %s", diff)
+        # If ID changed from `None to value we just created it no need to update
+        # diff is `(('add', ('spec', 'id'), None, 'R3JvdXA6MjAxNjc4OQ=='),)`
+        if (
+            len(diff) == 1
+            and diff[0][0] == "add"
+            and diff[0][1] == ("spec", "id")
+            and diff[0][2] is None
```
minhtule, line 27: nit: maybe
```python
if len(diff) == 1 and diff[0][:3] == ("add", ("spec", "id",), None):
```
like in the [connector handler](https://github.com/Twingate/kubernetes-operator/blob/a98acc3f96e7a03d3298e3d94c959ef991f1d821/app/handlers/handlers_connectors.py#L166)

app/handlers/handlers_groups.py
```diff
@@ -0,0 +1,71 @@
...
+@kopf.on.create("twingategroup")
+@kopf.on.update("twingategroup")
+def twingate_group_create_update(body, spec, logger, memo, patch, **kwargs):
+    logger.info("twingate_group_reconciler: %s", spec)
+    settings = memo.twingate_settings
+    client = TwingateAPIClient(settings)
+
+    if diff := kwargs.get("diff"):
+        logger.info("Diff: %s", diff)
+        # If ID changed from `None to value we just created it no need to update
```
minhtule, line 21: nit: missing backtick in <code>\`None</code>

app/handlers/handlers_groups.py
```diff
@@ -0,0 +1,71 @@
...
+    group_id = crd.spec.id
+    if group_id:
+        logger.info("Updating group with name='%s'", crd.spec.name)
+        try:
+            client.group_update(crd.spec)
+        except GraphQLMutationError as gqlerr:
+            logger.error("Failed to update group: %s", gqlerr)
+            if "does not exist" in gqlerr.message:
+                patch.spec["id"] = None
+            return fail(error=gqlerr.message)
```
minhtule, line 41: I wonder if we should do this in the client and throw some custom exception like `NotFound`. We could do this for other entities too?

minhtule (approved):
Some small suggestions. Fix & merge!

app/handlers/handlers_groups.py
ekampf: Maybe...  but if we do its its own PR - design change for client API.

ekampf: Added issue #400

PR #415, opened 2024-09-24
test: Improve integration tests stability

ekampf:
## Changes

- Fail fast on integration test failure


tests_integration/test_connector_flows.py
```diff
@@ -176,12 +152,12 @@ def test_connector_flows_image_change(run_kopf, random_connector_name):
...
-        time.sleep(10)
-        pod = kubectl_get("pod", connector_name)
+        time.sleep(5)
+        pod = kubectl_wait_pod_running(connector_name)
         assert pod["status"]["phase"] == "Running"
         assert pod["spec"]["containers"][0]["image"] == "twingate/connector:1.63.0"
 
-        kubectl_delete(f"tc/{connector_name}")
+        kubectl_delete_wait("tc", connector_name)
         time.sleep(10)
```
minhtule, line 161: We can delete this `sleep` now? 🤔

tests_integration/test_connector_flows.py
```diff
@@ -116,18 +99,17 @@ def test_connector_flows(run_kopf, random_connector_name):
                 }
             },
         )
-        time.sleep(10)
-        pod = kubectl_get("pod", connector_name)
+        time.sleep(5)
```
minhtule, line 102: I think it's nicer to delete this and let the retry logic in `kubectl_wait_pod_running` to sleep if neccessary.

tests_integration/test_connector_flows.py
```diff
@@ -176,12 +152,12 @@ def test_connector_flows_image_change(run_kopf, random_connector_name):
         # Change image tag
         # kubectl patch tc/test-connector-image-local -p '{"spec": {"image": {"tag": "1.63.0"}}}' --type=merge
         kubectl_patch(f"tc/{connector_name}", {"spec": {"image": {"tag": "1.63.0"}}})
-        time.sleep(10)
-        pod = kubectl_get("pod", connector_name)
+        time.sleep(5)
```
minhtule, line 155: Ditto

tests_integration/test_connector_flows.py
```diff
@@ -208,39 +184,31 @@ def test_connector_flows_pod_gone_while_operator_down(run_kopf, random_connector
...
-
         # pod was recreated
-        pod = kubectl_get("pod", connector_name)
+        pod = kubectl_wait_pod_running(connector_name)
         assert pod["status"]["phase"] == "Running"
 
         # Test done, delete connector
-        kubectl_delete(f"tc/{connector_name}")
+        kubectl_delete("tc", connector_name)
         time.sleep(5)
```
minhtule, line 205: Should we also use `kubectl_delete_wait` here?

tests_integration/test_connector_flows.py
```diff
@@ -286,14 +248,10 @@ def test_connector_flows_pod_migration_from_older_pod_with_finalizers(
...
-            pod = kubectl_get("pod", connector_name)
-
+        pod = kubectl_wait_pod_running(connector_name)
         assert pod["metadata"]["annotations"]["twingate.com/connector-podspec-version"] == "v1"  # fmt: skip
         assert pod["status"]["phase"] == "Running"
 
         # Test done, delete connector
-        kubectl_delete(f"tc/{connector_name}")
+        kubectl_delete("tc", connector_name)
         time.sleep(5)
```
minhtule, line 257: Ditto

minhtule (approved):
I have some questions but LGTM 👍

tests_integration/test_connector_flows.py
ekampf: No, if you patch the get the pod right after it might still be running - give k8s time to apply

tests_integration/test_connector_flows.py
ekampf: No, we delete the TwinGateConnector object but it takes time for k8s to propogate the change to its child obejects and delete them too

tests_integration/test_connector_flows.py
ekampf: no because after `tc` deletion is done we still need k8s to delete the secret and pod and `kubectl_delete_wait` wont do that

tests_integration/test_connector_flows.py
ekampf: same as above

tests_integration/test_connector_flows.py
ekampf: we patch the `tc` object and then its reconciler needs to run and patch the pod and only then pod restarts

PR #456, opened 2024-11-08
Fix  clusterrole with permissions over twingategroups objects

l-rossetti:

- Issue: TwingateGroup objects cannot be handled by the operator due to missing permissions over the TwingateGroup apis

## Changes

- Fix operator permissions over twingategroups apis

## Steps to reproduce:
- apply a twingate group object kind
- operator logs an error about missing permissions to list twingategroups
- the group is not created on twingate side


ekampf (approved):

ekampf:
Thanks!   releasing 0.12.1 for this

PR #487, opened 2024-12-13
feat: Enable Configuration of Kopf Watch Settings via Environment Variables

3schwartz:

We have observed that our operator stops listening to resource changes after running for some time.

After investigating the issue, we found several related Kopf issues that describe this behavior:
- https://github.com/nolar/kopf/issues/957
- https://github.com/nolar/kopf/issues/585

One solution would be to set the following:
```
settings.watching.client_timeout = 60
settings.watching.server_timeout = 60
```
However, passing these configurations to Kopf is currently not supported in the operator.

## Changes

Enabled configuration of Kopf’s watching settings through environment variables.

## Notes
- Kopf configuration: https://kopf.readthedocs.io/en/stable/configuration/


app/settings.py
```diff
@@ -27,6 +28,28 @@ def validate_graphql_global_id(value: str) -> str:
 GlobalID = Annotated[str, AfterValidator(validate_graphql_global_id)]
 
 
+class KopfWatchingSettings(BaseSettings):
+    model_config = SettingsConfigDict(env_prefix="KOPF_WATCHING_")
```
ekampf, line 32: I wonder if this should be `TWINGATE_WATCHING_` to be specific to this operator (vs. other kopf operators one might be running - might be confusing)
wdyt?

ekampf: or maybe just fold this into `TwingateOperatorSettings` by adding:

kopf_watching_server_timeout: float | None = None
kopd_watching_client_timeout: float | None = None
kopd_watching_connect_timeout: float | None = None
kopd_watching_reconnect_backoff: float | None = None

(which would map to `TWINGATE_KOPF_WATCHING_*` variables)

ekampf:
One more question - should we maybe default the configuration to what you suggested? (60)

app/settings.py
3schwartz: refactored to `TwingateOperatorSettings`

3schwartz:
> One more question - should we maybe default the configuration to what you suggested? (60s server&client timeout)

Based on a comment from the author of Kopf, it appears that setting a default value for this is not straightforward:
https://github.com/nolar/kopf/issues/585#issuecomment-735915945

The same point is reinforced in the Kopf documentation:
https://kopf.readthedocs.io/en/stable/configuration/#networking-timeouts
> The server-side timeouts are unpredictable, they can be 10 seconds or 10 minutes. Yet, it feels wrong to assume any “good” values in a framework (especially since it works without timeouts defined, just produces extra logs).

Given that this operator relies on Kopf, I would expect it to “behave the same” as Kopf’s default setup from a user’s perspective.

Of course, this is just my take on the matter :-)

ekampf (approved):

PR #491, opened 2024-12-16
fix: Allow extra env vars in the operator's Chart

ekampf:

Fixes https://github.com/Twingate/kubernetes-operator/pull/487

## Changes

We've added the feature to allow configuring KOPF using env vars but we also need the chart to support defining custom env vars

### DIFF
--- a/deploy/twingate-operator/values.yaml
+++ b/deploy/twingate-operator/values.yaml
@@ -74,10 +74,10 @@ affinity: {}
 
 priorityClassName: ""
 
-## Array with extra environment variables to add to the operator container
-## e.g:
-## extraEnvVars:
-##   - name: FOO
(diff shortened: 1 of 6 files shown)


deploy/twingate-operator/values.yaml
```diff
@@ -73,3 +73,11 @@ tolerations: []
 affinity: {}
 
 priorityClassName: ""
+
+## Array with extra environment variables to add to the operator container
+## e.g:
+## extraEnvVars:
+##   - name: FOO
+##     value: "bar"
+##
```
minhtule, line 82: nit: should be single `#` each line?

minhtule (approved):
LGTM 👍

Issue #530, opened 2025-01-24
Service annotation updates aren't handled


ekampf:
When updating resource annotations on a `Service` those updates arent being picked up by the operator
due to an issue with how `AnnotationsDiffBaseStorage` calculates the diff

Currently ugly workaround is to also change a non `twingate.com` annotation to make the operator work.

ekampf:
Solution for this would be:
1. We keep `kopf` state management on `twingate.com` prefix
2. We change Service annotations to be on `resource.twingate.com` prefix:
```
    resource.twingate.com: "true"
    resource.twingate.com/alias: myappchanged.internal
    resource.twingate.com/isBrowserShortcutEnabled: "false"
    resource.twingate.com/isVisible: "true"
    resource.twingate.com/name: My Service

```

PR #531, opened 2025-01-24
fix: Service annotations update not working

ekampf:

- Issue: #530 

## Changes

- Added support for new annotations on service, going forward these will be used and we'll deprecate the old ones:
```
    resource.twingate.com: "true"
    resource.twingate.com/alias: "myappchanged.internal"
    resource.twingate.com/isBrowserShortcutEnabled: "false"
    resource.twingate.com/isVisible: "true"
    resource.twingate.com/name: "My Service"
```


tests_integration/test_service_flows.py
```diff
@@ -13,6 +13,118 @@
...
+                "tcp": {
+                    "policy": "RESTRICTED",
+                    "ports": [{"end": 80, "start": 80}, {"end": 443, "start": 443}],
+                },
+                "udp": {"policy": "RESTRICTED", "ports": [{"end": 22, "start": 22}]},
+            },
+        }
+
+        # Test patching the service annotations updates the resource
+        kubectl_patch(
```
ekampf, line 101: this diff is a bit hard to read.
Basically I duplicated the service test.
The old test remains the same and the new test is the same + checkig annotaton change which is this code (lines 101-114)

liorr (approved):
Fix description and GTM

Issue #556, opened 2025-02-21
Chart image.tag references latest


flah00:
**What is missing?**

I noticed that the helm chart doesn't set Chart.yaml option `appVersion`. And values.yaml sets `image.tag` to `latest`. 

If we rely on latest we may have different operators running different versions... which could yield inconsistent results.

**Why do we need it?**

Setting `appVersion` gives you a few things for free
* The `appVersion` can be used to populate `image.tag`
  * See templates generated by `helm create`
* `appVersion` is also used to set the label `app.kubernetes.io/version`
  * `app.kubernetes.io/version` can be used to populate the datadog USM label `tags.datadoghq.com/version` [datadog universal service monitoring](https://docs.datadoghq.com/universal_service_monitoring/)
* `helm ls` displays app version, which can be helpful 

**Environment**

Any

**Twingate Kubernetes Operator version**

Any

**Anything else we need to know?**:

ekampf:
@flah00 thanks for the feedback!
Fix is merged and I will release new operator ASAP this week

ekampf:
`v0.17.0` is out

PR #560, opened 2025-02-24
feat: Add app-version to chart when releasing prod

ekampf:

- Issue: #556 

## Changes

Add appVersion to chart when releasing prod


romankor (approved):
GTM

Issue #581, opened 2025-03-21
fix: Support removing security policy from an access edge


minhtule:
**What is missing?**

For `TwingateResourceAccess` object, when the security policy is set to policy A and then unset (removed from the yaml file), the security policy A is not removed from the access. We should send `securityPolicy=null` in the `resourceAccessAdd` mutation.

**Why do we need it?**

Better experience

PR #583, opened 2025-03-21
fix: Support removing security policy from an access edge

ekampf:

- Issue: #581 

## Changes

- When calling the update mutation if `securityPolicyId` is missing it means "dont change whats on server. We should send `null` when the object is missing one to use the default


minhtule (approved):
LGTM 👍

PR #592, opened 2025-03-31
feat: Improve logs for better debugging experience

ekampf:
## Changes

- Propogate `logger` from handler down to the rest of the stack
- Add `handler` to all logs so we can filter for logs of a specific handler


minhtule (approved):
LGTM! 👍

PR #603, opened 2025-04-08
chore: Remove redundant integration test marker

clement0010:
## Changes
- Remove `integration` test marker on `TestConnectorCRD`


ekampf (changes requested):
Why is this redundant?   these tests need minikube...

ekampf (approved):

PR #605, opened 2025-04-09
feat: Support Twingate Resource tagging

clement0010:
## Changes
- Support adding k8s labels as Twingate Resource tags for `TwingateResource` CRD
- Since k8s `Service` also syncs the same `metadata.labels` to `TwingateResource` CRD  (see [code](https://github.com/Twingate/kubernetes-operator/blob/e6cd06849d5e958be082c97706e25a04374272ec/app/handlers/handlers_services.py#L86)), Twingate Resource tags will also be added as a result.

## Usage Example
```yaml
kind: TwingateResource
metadata:
  name: my-twingate-resource
  labels:
    env: prod # 👈🏼 This will add a `env:prod` tag in Twingate Resource
spec:
  name: My K8S Resource
  address: my.default.cluster.local
  alias: mine.local
```


app/crds.py
```diff
@@ -128,6 +141,7 @@ class ResourceSpec(BaseModel):
     is_visible: bool = True
     is_browser_shortcut_enabled: bool = True
     protocols: ResourceProtocols = Field(default_factory=ResourceProtocols)
+    tags: list[ResourceTag] = Field(default_factory=list, exclude=True)
```
clement0010, line 144: Technically, `tags` is not part of [`TwignateResource.spec`](https://github.com/Twingate/kubernetes-operator/wiki/API-Reference#twingateresourcespec), so we should exclude this field by default during model export.

app/handlers/handlers_resource.py
```diff
@@ -91,8 +95,8 @@ def twingate_resource_sync(spec, status, memo, logger, patch, **kwargs):
         else:
             # Resource was deleted, recreate it
             logger.info("Resource %s was deleted, recreating...", resource_id)
-            crd_withoput_id = crd.model_copy(update={"id": None})
-            resource = client.resource_create(crd_withoput_id)
+            crd_without_id = crd.model_copy(update={"id": None})
+            resource = client.resource_create(crd_without_id)
```
clement0010, line 99: fix typo

app/handlers/tests/test_handlers_resource.py
```diff
@@ -236,13 +265,16 @@ def test_sync_when_resource_exists_and_doesnt_need_update(
...
+        mock_api_client.get_resource.return_value = resource
+
         logger_mock = MagicMock()
         memo_mock = MagicMock()
         patch_mock = MagicMock()
         patch_mock.spec = {}
 
         twingate_resource_sync(
-            resource_spec.model_dump(by_alias=True, exclude="id"),
+            resource_spec.model_dump(by_alias=True),
```
clement0010, line 276: Since this is testing sync existing resource, we shouldn't exclude `id` here. Currently, the test will always pass because we skip the whole method ([here](https://github.com/Twingate/kubernetes-operator/blob/636c4961091cdd9e8deed6c69625a878ba417aa9/app/handlers/handlers_resource.py#L86)) if the `crd.id` is missing. 

In addition, the test setup is not complete, we should also return a resource in `get_resource`. This is fixed in `line 268` above

app/crds.py
clement0010: I updated the implementation, let me know what you think?

minhtule (approved):
LGTM 👍

tests_integration/test_service_flows.py
```diff
@@ -47,7 +50,10 @@ def test_service_flows(run_kopf, random_name_generator):
         tgr = kubectl_wait_object_handler_success(
             "twingateresource", resource_name, "twingate_resource_create"
         )
-
+        assert tgr["metadata"]["labels"] == {
+            "env": "dev",
+            "team": "engineering",
+        }
```
clement0010, line 56: @minhtule fyi, this was added after your review

ekampf (approved):

PR #606, opened 2025-04-10
feat: Support `syncLabels` on `TwingateResource.spec` and k8s `Service` annotations

clement0010:
## Changes
- Add `syncLabels` field on `TwingateResource.spec` to support whether to sync k8s metadata labels to Twingate Resource as tags. The default value is `True`
- Add `syncLabels` Twingate Resource annotations to k8s Service

## Notes
~⚠️ This PR is branched out from #605~


app/api/client_resources.py
```diff
@@ -266,10 +266,14 @@ def resource_create(
...
-                    for key, value in k8s_metadata.labels.items()
-                ],
+                "tags": (
+                    [
+                        {"key": key, "value": value}
+                        for key, value in k8s_metadata.labels.items()
+                    ]
+                    if resource.sync_labels
+                    else []
+                ),
```
clement0010, line 276: I am missing a test here, we should probably mock `execute_mutation` then assert it is called with empty `list`. I will add it tomorrow

app/crds.py
```diff
@@ -129,6 +129,7 @@ class ResourceSpec(BaseModel):
     is_visible: bool = True
     is_browser_shortcut_enabled: bool = True
     protocols: ResourceProtocols = Field(default_factory=ResourceProtocols)
+    sync_labels: bool = Field(default=True, exclude=True)
```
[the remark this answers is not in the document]
clement0010: We use `ResourceSpec.model_dump()` in test ([example](https://github.com/Twingate/kubernetes-operator/blob/bcc2d494b106abbc34268cd004d23a2de5d76dfb/app/api/tests/test_client_resources.py#L192-L199)), I wanted to exclude it by default so we don't do `model_dump(exclude=["sync_labels"])` everywhere in the test.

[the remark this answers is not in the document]
minhtule: Ahh, I see. We shouldn't set default behaviors (or values) just based on usages in test.

app/api/tests/test_client_resources.py
```diff
@@ -190,7 +190,53 @@ def test_resource_create(
...
+    def test_resource_create_with_disable_sync_labels(
+        self, test_url, api_client, resource_factory, mocked_responses
+    ):
+        resource = resource_factory(tags=[])
+        crd = resource.to_spec(id=None, sync_labels=False)
+        k8s_metadata = K8sMetadata(
+            name="my-resource",
+            namespace="default",
+            uid="ad0298c5-b84f-4617-b4a2-d3cbbe9f6a4c",
+            labels=resource.to_metadata_labels(),
```
minhtule, line 215: This `resource` instance has no tags so this label is empty right? So regardless of `sync_labels` value, the mutation is called with an empty list of tags?

app/api/tests/test_client_resources.py
```diff
@@ -257,7 +307,54 @@ def test_resource_update(
...
+    def test_resource_update_with_disable_sync_labels(
+        self, test_url, api_client, resource_factory, mocked_responses
+    ):
+        resource = resource_factory(tags=[])
+        crd = resource.to_spec(sync_labels=False)
+        k8s_metadata = K8sMetadata(
+            name="my-resource",
+            namespace="default",
+            uid="ad0298c5-b84f-4617-b4a2-d3cbbe9f6a4c",
+            labels=resource.to_metadata_labels(),
```
minhtule, line 332: Ditto

minhtule (approved):
I have some questions on the unit tests. Otherwise, LGTM 👍

ekampf (approved):

app/api/tests/test_client_resources.py
clement0010: AH yes, you're right. We shouldn't have set empty labels here. I updated the test in this commit [1933533](https://github.com/Twingate/kubernetes-operator/pull/606/commits/193353348aac106aaaa82914081c9f2a25c7ff8e). The test should fail when we enable `syncLabels`.

Issue #604, opened 2025-04-08
feat: Update TwingateConnector with liveness\readiness probes and readOnlyRootFilesystem


twingate-blee:
**What is missing?**
We want to be able to set the securityContext  with `readOnlyRootFilesystem: true`

We want to add livenessProbe and readinessProbe

**Why do we need it?**
Have `readOnlyRootFilesystem` will help the security posture
Adding livenessProbe and readinessProbe will help with detecting pod issues.

These changes will match the settings for the connector helm chart

**Anything else we need to know?**:
To support `readOnlyRootFilesystem: true` we will need to add a emptyDir mount for /var/run/twingate/connector.sock

```
        volumeMounts:
        - name: twingate-socket
          mountPath: /var/run/twingate
      volumes:
      - name: twingate-socket
        emptyDir: {}
```

ekampf:
Resolved in https://github.com/Twingate/kubernetes-operator/pull/612

PR #612, opened 2025-04-15
feat: Support `livenessProbe`, `readinessProbe` and enhance security by setting `readOnlyRootFilesystem` on `TwingateConnector`

clement0010:

- Issue: https://github.com/Twingate/kubernetes-operator/issues/604

## Changes
- Add `livenessProbe` and `readinessProbe` to `TwingateConnector.spec.container`
- Set `readOnlyRootFilesystem` in `securityContext` to `True` by default. Also, add `volumes` and `volumeMounts`

### DIFF
--- a/app/handlers/handlers_connectors.py
+++ b/app/handlers/handlers_connectors.py
@@ -47,6 +47,8 @@ def get_connector_pod(
     container_extra = spec.container_extra
     extra_env = container_extra.pop("env", [])
     extra_env_from = container_extra.pop("envFrom", [])
+    extra_volumes = container_extra.pop("volumes", [])
+    extra_volume_mounts = container_extra.pop("volumeMounts", [])
 
     # fmt: off
     pod_spec = {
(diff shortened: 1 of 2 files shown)


app/handlers/handlers_connectors.py
```diff
@@ -74,11 +74,39 @@ def get_connector_pod(
...
+                        "name": "twingate-socket",
+                        "mountPath": "/var/run/twingate",
+                    }
+                ],
                 **container_extra,
             },
             *spec.sidecar_containers,
         ],
+        "volumes": [{"name": "twingate-socket", "emptyDir": {}}],
         **spec.pod_extra,
```
minhtule, line 110: I'm not 100% sure about this but I think we should pop the `volumes` key from `spec.pod_extra` dict and `volumeMounts` from `spec.container_extra` because a user might specify those keys and unintentionally break the health checks when `readOnlyRootFilesystem` is true. @ekampf WDYT?

ekampf: We need to support `pod_extra` having other volumes

app/handlers/tests/test_handlers_connector.py
```diff
@@ -395,3 +401,46 @@ def test_not_past_due_annotation(
         assert run.patch_mock.meta == {}
         run.k8s_client_mock.patch_namespaced_pod.assert_not_called()
         mock_get_image.assert_not_called()
+
+
+class TestGetConnectorPod:
```
clement0010, line 406: It'd be nice to cover other spec fields logic as well for completeness. I can do it in a separate PR.

ekampf (approved):

app/handlers/handlers_connectors.py
```diff
@@ -47,6 +47,8 @@ def get_connector_pod(
     container_extra = spec.container_extra
     extra_env = container_extra.pop("env", [])
     extra_env_from = container_extra.pop("envFrom", [])
+    extra_volumes = container_extra.pop("volumes", [])
```
minhtule, line 50: `volumes` field should be specified in `pod_extra` , not `container_extra` like `volumeMounts` ([example](https://kubernetes.io/docs/concepts/storage/volumes/#configmap))

minhtule (changes requested):

minhtule (approved):
LGTM 👍

PR #613, opened 2025-04-15
feat: Supporting adding default tags on Twingate Resource

clement0010:
## Changes
- Add default tags to Twingate Resource based on `TWINGATE_DEFAULT_RESOURCE_TAGS` envvar set on the Operator.
- Refactor Resource client API code

### DIFF
--- a/app/api/client_resources.py
+++ b/app/api/client_resources.py
@@ -259,9 +259,8 @@ class TwingateResourceAPIs:
         is_browser_shortcut_enabled: bool,
         remote_network_id: str,
         security_policy_id: str | None,
-        protocols: ResourceProtocols,
+        protocols: dict[str, Any],
         tags: list[dict[str, str]],
-        **_kwargs,
     ) -> Resource:
         result = self.execute_mutation(
             "resourceCreate",
@@ -293,7 +292,6 @@ class TwingateResourceAPIs:
         security_policy_id: str | None,
         protocols: ResourceProtocols,
         tags: list[dict[str, str]],
-        **_kwargs,
     ) -> Resource | None:
         result = self.execute_mutation(
             "resourceUpdate",
--- a/app/api/client_resources.py
+++ b/app/api/client_resources.py
@@ -8,7 +8,7 @@ from pydantic.alias_generators import to_camel
 
 from app.api.exceptions import GraphQLMutationError
 from app.api.protocol import TwingateClientProtocol
-from app.crds import K8sMetadata, ProtocolPolicy, ResourceSpec
+from app.crds import ProtocolPolicy, ResourceSpec
 
 
 class ResourceAddress(BaseModel):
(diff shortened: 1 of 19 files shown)


app/handlers/handlers_resource.py
```diff
@@ -3,20 +3,27 @@
...
+    logger.info("Got a create request: %s. Labels: %s", spec, labels)
     resource = ResourceSpec(**spec)
     client = TwingateAPIClient(memo.twingate_settings, logger=logger)
+    labels = {
+        **{
+            tag["name"]: tag["value"]
+            for tag in memo.twingate_settings.default_resource_tags
+        },
+        **dict(labels),
+    }
```
clement0010, line 21: Currently, if there are duplicate tags. The labels on Resource CRD will take precedence and overwrite the default tag

app/handlers/handlers_resource.py
```diff
@@ -3,20 +3,27 @@
...
+    graphql_arguments = resource.to_graphql_arguments(labels=labels)
 
     # Support importing existing resources - if `id` already exist we assume it's already created
     if resource.id:
-        resource = client.resource_update(resource, k8s_metadata)
+        resource = client.resource_update(**graphql_arguments)
```
clement0010, line 26: Another alternative we discussed was to use [`client_resource.Resource`](https://github.com/Twingate/kubernetes-operator/blob/9fb816874e298b42c7b574f734bb800a05fabc29/app/api/client_resources.py#L66-L82), which represents a Twingate Resource. This is probably not a good idea because we don't have all the fields like `id`, `created_at` to construct the model instance.

Another option is to create a `ResourceArguments` pydantic model with fields as the Resource graphQL mutation arguments. Then we convert `ResourceSpec` to `ResourceArguments`. I think this option is overkill and would end up being very similar to the current implementation.

app/handlers/handlers_resource.py
ekampf: yes thats the expected behavior 👍

app/handlers/handlers_resource.py
```diff
@@ -3,20 +3,27 @@
...
 @kopf.on.create("twingateresource")
-def twingate_resource_create(body, meta, spec, memo, logger, patch, **kwargs):
-    logger.info("Got a create request: %s. Metadata: %s", spec, meta)
-    k8s_metadata = K8sMetadata(**meta)
+def twingate_resource_create(body, labels, spec, memo, logger, patch, **kwargs):
+    logger.info("Got a create request: %s. Labels: %s", spec, labels)
     resource = ResourceSpec(**spec)
     client = TwingateAPIClient(memo.twingate_settings, logger=logger)
+    labels = {
+        **{
```
ekampf, line 16: nit: I'd put this in the settings class as `memo.twingate_settings.default_resource_tags_dict`

ekampf: I also like the `|` opeartion better than ** creation:
```
memo.twingate_settings.default_resource_tags_dict | dict(labels)
```

ekampf (approved):
Fix and Merge

app/handlers/tests/test_handlers_resource.py
```diff
@@ -29,26 +30,48 @@ def mock_k8s_metadata():
...
+    return MagicMock(
+        twingate_settings=TwingateOperatorSettings(
+            network="slug",
+            host="test.com",
+            api_key="test_key",
+            remote_network_id="UmVtb3RlTmV0d29yazoxMjMK",
+            default_resource_tags=[
+                {
+                    "name": "default",
+                    "value": "value",
```
minhtule, line 44: nit: The `"value": "value"` is not a good example. Let's use something like
```python
{
    "name": "env",
    "value": "test",
}
```

minhtule (approved):
Fix & merge!

app/handlers/handlers_resource.py
minhtule: Yeah, there are a few other places that could be simplified using dict union `|` operator. We should update them too.

app/api/client_resources.py
```diff
@@ -250,57 +250,65 @@ def get_resource(self: TwingateClientProtocol, resource_id: str) -> Resource | N
...
+        self: TwingateClientProtocol,
+        *,
+        name: str,
+        address: str,
+        alias: str | None,
+        is_visible: bool,
+        is_browser_shortcut_enabled: bool,
+        remote_network_id: str,
+        security_policy_id: str | None,
+        protocols: ResourceProtocols,
```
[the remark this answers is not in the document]
clement0010: Yes the type should dict here

app/api/client_resources.py
```diff
@@ -250,57 +250,63 @@ def get_resource(self: TwingateClientProtocol, resource_id: str) -> Resource | N
...
+        *,
+        id: str,
+        name: str,
+        address: str,
+        alias: str | None,
+        is_visible: bool,
+        is_browser_shortcut_enabled: bool,
+        remote_network_id: str,
+        security_policy_id: str | None,
+        protocols: ResourceProtocols,
```
minhtule, line 293: @clement0010 We should also update the typing here like in the `resource_create()` method.

app/handlers/handlers_resource.py
```diff
@@ -3,20 +3,24 @@
...
-    logger.info("Got a create request: %s. Metadata: %s", spec, meta)
-    k8s_metadata = K8sMetadata(**meta)
+def twingate_resource_create(body, labels, spec, memo, logger, patch, **kwargs):
+    logger.info("Got a create request: %s. Labels: %s", spec, labels)
     resource = ResourceSpec(**spec)
     client = TwingateAPIClient(memo.twingate_settings, logger=logger)
+    labels = {
+        tag["name"]: tag["value"]
+        for tag in memo.twingate_settings.default_resource_tags
+    } | dict(labels)
```
minhtule, line 18: Why don't we move this conversation 
```
{
    tag["name"]: tag["value"]
    for tag in memo.twingate_settings.default_resource_tags
}
```
to `TwingateOperatorSettings` as [previously suggested](https://github.com/Twingate/kubernetes-operator/pull/613#discussion_r2045036542)? 🤔 It's repeated 3 times here.

app/handlers/tests/test_handlers_resource.py
```diff
@@ -62,7 +90,10 @@ def test_create(
 
         mock_api_client.resource_update.assert_not_called()
         mock_api_client.resource_create.assert_called_once_with(
-            resource_spec, K8sMetadata(**mock_k8s_metadata)
+            **resource_spec.to_graphql_arguments(
+                labels={"managed_by": "test"} | mock_k8s_metadata["labels"],
```
minhtule, line 94: It's clearer to explicitly layout the expected value here
```python
labels = {"manage_by": "test", "env": "test"}
```

app/handlers/handlers_resource.py
minhtule: We could keep `default_resource_tags` Pydantic field definition as is and create a property that derives from the field e.g.
```python
class TwingateOperatorSettings(BaseSettings):
    default_resource_tags: list[dict[str, str]] = []

    @property
    def default_resource_k8s_labels() -> dict:
        return {
            tag["name"]: tag["value"]
            for tag in self.default_resource_tags
        }
```

PR #617, opened 2025-04-16
fix: `defaultResourceTags` to match labels schema

ekampf:
## Changes

Changed `defaultResourceTags` schema in `values.yaml` to match how labels are defined in k8s - as an object rather than a list of objects. (see example: https://github.com/instrumenta/kubernetes-json-schema/blob/master/v1.7.8/_definitions.json#L528)


minhtule (approved):
LGTM! 👍

PR #622, opened 2025-04-18
fix: TWINGATE_DEFAULT_RESOURCE_TAGS default value should be empty dict

ekampf:
## Changes

- TWINGATE_DEFAULT_RESOURCE_TAGS default value should be empty dict


minhtule (approved):
LGTM 👍

PR #623, opened 2025-04-18
fix: Operator should halt if settings are invalid

ekampf:
## Changes

- Raise `RuntimeException` if failed to load settings


minhtule (approved):
LGTM 👍

PR #634, opened 2025-05-07
fix: Add delete permission to Twingate resources

ekampf:
## Changes

Added `delete` permission to twingate resources.
When removing resource annotation from a `Service` object the operator needs to delete the corresponding `TwingateResource` object


minhtule (approved):
LGTM 👍

PR #635, opened 2025-05-07
chore: Update coveralls dependency

ekampf:
## Changes

Updated `coveralls` and `pytest-cov`


pyproject.toml
```diff
@@ -7,7 +7,7 @@ readme = "README.md"
 package-mode = false
 
 [tool.poetry.dependencies]
-python = "^3.11"
```
ekampf, line 10: Coveralls defined that it doesnt support Python 3.13 (`<3.13`) and since `^3.11` does support it it prevented from upgrade

minhtule (approved):
LGTM 👍

PR #636, opened 2025-05-08
fix: update operator ClusterRole permissions for twingate.com apiGroup

ekampf:
## Changes

- Operator should have full control on `twingate.com` resources. This will make us future proof as we add features and resources


minhtule (approved):
LGTM 👍

PR #640, opened 2025-05-12
test: Remove unused `pytest-freezegun` library

ekampf:
## Changes

Removed the unused `pytest-freezegun` library.
In any case, if we ever do need this functionality, we should be using `pendulum`'s testing functions:
https://pendulum.eustace.io/docs/#testing


minhtule (approved):
LGTM! 👍

PR #654, opened 2025-05-22
fix: Reconciler should only update k8s if there’s drift in image

ekampf:
## Changes

Reconciler was always calling k8s replace because image if statement was wrong.


minhtule (approved):
LGTM 👍

PR #664, opened 2025-05-28
fix: allow_zero_version should be true

ekampf:
## Changes

https://python-semantic-release.readthedocs.io/en/latest/upgrading/10-upgrade.html#default-configuration-changes

Default used to be `true` and changed to `false` in v10


minhtule (approved):
LGTM!

PR #668, opened 2025-06-02
feat: Support creating `LoadBalancer` Service type for Kubernetes Resource

clement0010:
## Changes
- Create `proxy.address` with the ingress IP address when Service type is `LoadBalancer`
- Retry `twingate_service_create` handler when waiting to provision the IP address

### DIFF
--- a/app/handlers/handlers_services.py
+++ b/app/handlers/handlers_services.py
@@ -1,9 +1,10 @@
 import base64
 from collections.abc import Callable
+from enum import StrEnum
 
 import kopf
 import kubernetes
-from kopf import Body
+from kopf import Body, Status
 
 from app.crds import ResourceType
 from app.handlers import success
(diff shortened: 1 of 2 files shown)


app/handlers/handlers_services.py
```diff
@@ -70,10 +71,30 @@ def get_ca_cert(tls_secret: kubernetes.client.V1Secret) -> str:
...
+    CLUSTER_IP = "ClusterIP"
+    LOAD_BALANCER = "LoadBalancer"
+
+
 def service_to_twingate_resource(service_body: Body, namespace: str) -> dict:
     meta = service_body.metadata
     spec = service_body.spec
-    service_name = service_body.meta.name
+    status = service_body.status
+    service_name = service_body.meta.name or ""
```
clement0010, line 97: This change is to fix mypy `error: Argument 2 to "validate_load_balancer_status" has incompatible type "str | None"; expected "str"  [arg-type]`

app/handlers/tests/test_handlers_services.py
```diff
@@ -240,7 +296,75 @@ def test_kubernetes_resource_type_annotation_with_invalid_ca_cert(
...
+    )
+    def test_retry_load_balancer_ip_is_not_ready(
+        self,
+        example_load_balancer_gateway_service_body,
+        kopf_handler_runner,
+        k8s_customobjects_client_mock,
+        status,
+    ):
+        # ruff: noqa: SLF001
+        example_load_balancer_gateway_service_body._status = status
```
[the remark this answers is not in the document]
minhtule: @clement0010 Can we patch the `Body.status` property instead of setting the protected attribute?

app/handlers/handlers_services.py
minhtule: I would avoid setting a variable value to fix a type issue. Instead I would consider
- Annotate `service_name: str | None` in `validate_load_balancer_status()` helper function
- Use `cast(str, service_body.meta.name)` to let mypy know `name` cannot be `None` here
- Pass the `service_body: Body` to the `validate_load_balancer_status`

clement0010: Lets go with the second option since, `service_name` cannot be `None` here

minhtule (approved):
LGTM! 👍

ekampf (approved):

PR #674, opened 2025-06-05
fix: Pod to Deployment migration doesnt work with imagePolicy

### LINKED ISSUE #673: When upgrading from 0.21.X -> 0.22.X pods deleted, but deployment not created until the TwingateConnector resource is updated
**What happened?**

When updating from version `0.21.2` of the twingate-operator to `0.22.1`, the operator came up and correctly deleted the pre-deployment _Pods_ (as expected) killing all connectors in the cluster.  However, new _Deployments_ were **not** created until we manually edited the _TwingateConnector_ resources (or created new ones).

I believe this is due to this method [`twingate_connector_resume`](https://github.com/Twingate/kubernetes-operator/blob/main/app/handlers/handlers_connectors.py#L214-L219) deleting the pre-deployment pods (as it should), but creating and updating deployments only occur on the Create or Update actions on a _TwingateResource_ instance.

**Did you expect to see something different?**
My expectation is the swap from pre-deployment _Pod_ to new _Deployment_ would happen behind the scenes without requiring "updating" existing _TwingateConnector_ resources.

**How to reproduce it (as minimally and precisely as possible)**:

Have a deployment of `0.21.2` twingate-operator running.

Create a connector:
```
kind: TwingateConnector
metadata:
  labels:
    app.kubernetes.io/name: twingate-connector
  name: twingate-connector-01
spec:
  hasStatusNotificationsEnabled: true
  imagePolicy:
    allowPrerelease: false
    provider: dockerhub
    repository: twingate/connector
    schedule: 0 4 * * *
  logLevel: 6
  podExtra:
    topologySpreadConstraints:
    - labelSelector:
        matchLabels:
          app.kubernetes.io/name: twingate-connector
      maxSkew: 1
      minDomains: 2
      topologyKey: topology.kubernetes.io/zone
      whenUnsatisfiable: DoNotSchedule
```

Update to `0.22.0` on the twingate-operator.

Notice: the pods will disappear (as expected) when the new version of the operator is ready, but the deployment will **NOT** appear.

Update (via `kubectl edit...` the `logLevel` in the TwingateConnector above to `5`, save exit.

Notice: The deployment will appear.

**Environment**

N/A

* Twingate Kubernetes Operator version: `0.22.1` from `0.21.2`

    `Insert image tag or Git SHA here`
    

* Kubernetes version information: `v1.32.3-eks-bcf3d70`

    `kubectl version`
    

* Kubernetes cluster kind: EKS

    insert how you created your cluster: kops, bootkube, etc.

* Manifests:

```
insert manifests relevant to the issue
```

* Twingate Kubernetes Operator Logs:

```
insert Twingate Kubernetes Operator logs relevant to the issue here
```

While its certainly not super impactful to update the _TwingateConnector_ resources in a cluster after the new operator version is running, there is a temporary breakage in access to that cluster as the connectors are "gone".

ekampf:

Fixes #673 

## Changes

When upgrading connectors from `Pod` to `Deployment`, the code path using `imagePolicy` was missing a check - it would only check the deployment when the policy's `schedule` is triggered which is usually daily or longer.
This change fixes `twingate_connector_pod_reconciler` to recreate the deployment if it doesnt exist, regardless of the `schedule` value

### DIFF
--- a/app/handlers/handlers_connectors.py
+++ b/app/handlers/handlers_connectors.py
@@ -335,7 +335,7 @@ def twingate_connector_pod_reconciler(
             meta.annotations.get(ANNOTATION_NEXT_VERSION_CHECK, "0001-01-01 00:00:00")
         )
 
-        if now >= next_check_at:
+        if now >= next_check_at or not k8s_deployment:
             patch.meta["annotations"] = {
                 ANNOTATION_LAST_VERSION_CHECK: now.to_iso8601_string(),
                 ANNOTATION_NEXT_VERSION_CHECK: crd.spec.image_policy.get_next_date_iso8601(),
--- a/app/handlers/tests/test_handlers_connector.py
+++ b/app/handlers/tests/test_handlers_connector.py
@@ -439,9 +439,6 @@ class TestTwingateConnectorPodReconciler_ImagePolicy:
             with_id=True,
         )
 
-        mock_dep = MagicMock()
-        k8s_apps_client_mock.read_namespaced_pod.return_value = mock_dep
-
         run = kopf_handler_runner(twingate_connector_pod_reconciler, crd, MagicMock())
         assert run.result == {"success": True, "ts": ANY}
         run.k8s_apps_client_mock.replace_namespaced_deployment.assert_called_once()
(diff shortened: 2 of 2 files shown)


minhtule (approved):
LGTM 👍

PR #681, opened 2025-06-11
fix: Only call resource udate when spec\labels update

ekampf:
## Changes

Optimize `twingate_resource_update` so that it doesnt call API to update the resource unless there was a change in `spec` o `metadata.labels`


minhtule (approved):
LGTM 👍

PR #685, opened 2025-06-12
fix: Can’t patch status of Service objects

ekampf:
## Changes

Resolve unecessary error logs of patching inconsistency:
```
Patching failed with inconsistencies: (('remove', ('status', 'twingate_service_create'), {'ts': '2025-06-12T01:53:08.362368', 'success': True, 'message': 'Created TwingateResource foo'}, None),)
```
When the service's handlers return a structure `kopf` will try to patch the `Service.status` and fail...

Reason is you can't patch k8s's  `Service.status`. Its strongly typed and doesnt allow new data.
Switched to using events (shown when describing the service instead)


minhtule (approved):
LGTM 👍

PR #704, opened 2025-06-27
feat: Improve validation - ‘spec’ is always required

ekampf:
We always assume `spec` is present in code. But the CRD allowd objects without `spec` which doesnt really make sense


minhtule (approved):
LGTM! 👍

PR #710, opened 2025-07-01
feat: Delete Kubernetes Access Gateway k8s Service object before removing operator

clement0010:
## Changes
- Add `pre-delete-cleanup.yaml` to delete Kubernetes Access Gateway's k8s Service object before Helm removes Twingate Operator

## Notes
When we do `helm uninstall <Operator release name>`, Twingate Operator is deleted before it can remove the Gateway's k8s Service object. As a result, the k8s Service object and its associated Twingate Resource CRD are left dangling.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -592,49 +592,49 @@ pytz = ">2021.1"
 
 [[package]]
 name = "cryptography"
-version = "45.0.4"
+version = "45.0.5"
 description = "cryptography is a package which provides cryptographic recipes and primitives to Python developers."
 optional = false
 python-versions = "!=3.9.0,!=3.9.1,>=3.7"
 groups = ["main"]
 files = [
-    {file = "cryptography-45.0.4-cp311-abi3-macosx_10_9_universal2.whl", hash = "sha256:425a9a6ac2823ee6e46a76a21a4e8342d8fa5c01e08b823c1f19a8b74f096069"},
-    {file = "cryptography-45.0.4-cp311-abi3-manylinux2014_aarch64.manylinux_2_17_aarch64.whl", hash = "sha256:680806cf63baa0039b920f4976f5f31b10e772de42f16310a6839d9f21a26b0d"},
-    {file = "cryptography-45.0.4-cp311-abi3-manylinux2014_x86_64.manylinux_2_17_x86_64.whl", hash = "sha256:4ca0f52170e821bc8da6fc0cc565b7bb8ff8d90d36b5e9fdd68e8a86bdf72036"},
-    {file = "cryptography-45.0.4-cp311-abi3-manylinux_2_28_aarch64.whl", hash = "sha256:f3fe7a5ae34d5a414957cc7f457e2b92076e72938423ac64d215722f6cf49a9e"},
-    {file = "cryptography-45.0.4-cp311-abi3-manylinux_2_28_armv7l.manylinux_2_31_armv7l.whl", hash = "sha256:25eb4d4d3e54595dc8adebc6bbd5623588991d86591a78c2548ffb64797341e2"},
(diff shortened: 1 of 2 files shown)


deploy/twingate-operator/templates/pre-delete-cleanup.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if index .Values "kubernetes-access-gateway" "enabled" }}
```
clement0010, line 1: We can't write `.Values.kubernetes-access-gateway.enabled` here because `kubernetes-access-gateway` is not a standard naming convention, it's a known [design flaw](https://github.com/helm/helm/issues/9636#issuecomment-1481749407) when using sub-chart. This is a workaround I found on this [thread](https://github.com/helm/helm/issues/2192#issuecomment-306033149).

deploy/twingate-operator/templates/pre-delete-cleanup.yaml
```diff
@@ -0,0 +1,27 @@
...
+      containers:
+        - name: cleanup
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: bitnami/kubectl:latest
+          command:
+            - /bin/sh
+            - -c
+            - |
+              kubectl delete svc -n {{ .Release.Namespace }} {{ include "gateway.fullname" (index .Subcharts "kubernetes-access-gateway") }}
```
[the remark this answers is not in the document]
ekampf: need to add `kubectl wait --for=delete ...` to make sure resource is deleted

ekampf (changes requested):
Missing tests

deploy/twingate-operator/templates/pre-delete-cleanup.yaml
clement0010: Yeah. Looking at the [standard library](https://pkg.go.dev/text/template#hdr-Functions) used in Helm, a map will work here too.

> index
	Returns the result of indexing its first argument by the
	following arguments. Thus "index x 1 2 3" is, in Go syntax,
	x[1][2][3]. Each indexed item must be a map, slice, or array.

clement0010: Btw, there’s a weird behavior that `dig` doesn’t work on .Values (see [issue](https://github.com/helm/helm/issues/9266)) e.g. dig "kubernetes-access-gateway" "enabled" "false" .Values gives this error:
```
executing "twingate-operator/templates/NOTES.txt" at <dig "kubernetes-access-gateway" "enabled" "false" .Values>: error calling dig: interface conversion: interface {} is chartutil.Values, not map[string]interface {}
```

The workaround is use `dig "kubernetes-access-gateway" "enabled" "false"  (.Values | dict )`

clement0010: The reason we are getting `dig: interface conversion: interface {} is chartutil.Values, not map[string]interface {}` is because of the type assertion [here](https://github.com/Masterminds/sprig/blob/8cb06fe3c8b0f1163c26b0a558669da72ee14656/dict.go#L154)

Go expected a type of `map[string]interface{}` but got `chartutil.Values` instead. Though `chartutil.Values` is just a wrapper of [`map[string]interface{}`](https://github.com/gabrtv/helm/blob/eee6e143d89c55d1f481d54ba2deb891ae60446f/pkg/chartutil/values.go#L37-L38), Go expects us to convert the type explicitly.

clement0010: I also checked `index` returns nil (evaluate to `false`) when `kubernetes-access-gateway.enabled` key is missing. On another note, we must use `index` in `line 2` because `dig` traverses another level down (`Subcharts.kubernetes-access-gateway.<additional_key>`) but we only needed first level (`Subcharts.kubernetes-access-gateway`).

minhtule (approved):
LGTM 👍

ekampf (approved):

PR #728, opened 2025-07-21
chore: Replace golden test with `helm unittest`

clement0010:
## Changes
- Replace golden test with `helm unittest` plugin


deploy/twingate-operator/tests/values.test.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+imagePullSecrets: []
+
+nameOverride: ""
+fullnameOverride: ""
+
+serviceAccount:
+  create: true
+  annotations: {}
+  name: ""
```
minhtule, line 27: It's unclear to me when I should add to this `values.test.yaml` to be captured by the snapshot test. These are default values. Some other values e.g. `podAnnotations` or `podLabels` are not. I think the snapshot test should test the default `values.yaml` file. Other custom fields like `podAnnotations` should have be tested on its own in the `deployment_test.yaml` because they only affects the deployment template.

deploy/test/golden/pre-delete-cleanup.golden.yaml
minhtule, line 1: Previously this test would also test that the `kubernetes-access-gateway` subchart is enabled correctly. We should have a test for that.

deploy/twingate-operator/tests/values.test.yaml
clement0010: I can see why it might be unclear. The reason I added `values.test.yaml` is so we don't have to keep adding the 3 required fields (`network`, `apiKey` and `remoteNetwork`) in `twingateOperator` for other test suites repeatedly. I updated the setup ([80a85d8](https://github.com/Twingate/kubernetes-operator/pull/728/commits/80a85d82721306451ff5827a4807242a88ced1c6)) to set the `twingateOperator` value at the test suite level instead.

deploy/twingate-operator/tests/__snapshot__/pre-delete-cleanup_test.yaml.snap
```diff
@@ -1,5 +1,158 @@
...
+      template:
+        metadata:
+          annotations:
+            checksum/tlsAlternativeNames: eee78e4712e57977935ceb9ebfdb354aa33c85ca4dbb490001827f64259b6b8d
+          labels:
+            app.kubernetes.io/instance: RELEASE-NAME
+            app.kubernetes.io/managed-by: Helm
+            app.kubernetes.io/name: kubernetes-access-gateway
+            app.kubernetes.io/version: 0.7.0
+            helm.sh/chart: kubernetes-access-gateway-0.7.0
```
minhtule: This snapshot test would fail in dependabot PR updating the Gateway subchart (e.g. https://github.com/Twingate/kubernetes-operator/pull/725). Is there a way to avoid having to manually fix those dependabot PR?

deploy/twingate-operator/tests/values.test.yaml
minhtule: If that's the issue, why do we need other values than those 3 required fields (`network`, `apiKey` and `remoteNetwork`) in the `values.test.yaml`?

deploy/twingate-operator/tests/deployment_test.yaml
```diff
@@ -84,3 +84,8 @@ tests:
     asserts:
       - failedTemplate:
           errorPattern: "twingateOperator: Must validate one and only one schema (oneOf)"
+  - it: should use optional values
+    values:
+      - values.test.yaml
+    asserts:
+      - matchSnapshot: {}
```
minhtule, line 91: Why don't we test optional values like the other tests in this file? I find this snapshot test quite problematic
- The `values.test.yaml` still have a mixed of default values and customized values. It's confusing to other maintainers what should go into that file.
- When I add a required values, I wouldn't expect to update this snapshot.

deploy/twingate-operator/tests/__snapshot__/pre-delete-cleanup_test.yaml.snap
clement0010: Yes, it was already fixed in this [commit](https://github.com/Twingate/kubernetes-operator/pull/728/commits/1f8bfacee713e95c3e2ea001fbe3f4ecfd89c7b6)

minhtule: That file is called [pre-delete-cleanup_test.yaml](https://github.com/Twingate/kubernetes-operator/pull/728#diff-d29f200b6c731eba23039df74be09f98ae28e74a85ccfc77c90b8a9ffb2188a3) but it tests 2 things: the subchart and the clean up job. I think they should be separately so when we make changes to the subchart, only that test should fail.

clement0010: Updated in [493c647](https://github.com/Twingate/kubernetes-operator/pull/728/commits/493c6477b0e8e2416804667ddb5047f156a701dc)

deploy/twingate-operator/tests/deployment_test.yaml
clement0010: Initially, I thought we might want to test other customized values like `podLabels`, `podAnnotations`, `nodeSelector`, etc which would make the test bloated. On second thought, i think we don't need to test these field since it's trivial. 

So we can just keep the test simple and only test `logFormat`, `logVerbosity`, `defaultResourceTags` and `extraEnvVars` which have some custom logic in `deployment.yaml`

deploy/twingate-operator/tests/snapshot_test.yaml
minhtule, line 1: We have snapshot tests across different test files so it's unclear what should go here. I would rename this to something more specific e.g. `default_values_test.yaml`

deploy/twingate-operator/tests/deployment_test.yaml
minhtule, line 1: I would split this suite into multiple ones to reuse a setup for each suite and it's clearer to other maintainers
- API key (provided value vs provided secret)
- Remote network (ID, name and other validation)
- Other fields in the deployment (pod labels, extra env var, etc)

deploy/twingate-operator/tests/deployment_test.yaml
```diff
@@ -0,0 +1,140 @@
...
+        remoteNetworkId: "<remote network id>"
+      extraEnvVars:
+        - name: FOO
+          value: bar
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: FOO
+            value: "bar"
```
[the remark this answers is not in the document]
minhtule: @clement0010 Why do we use double quote here but not in line 134?

deploy/twingate-operator/tests/deployment_test.yaml
clement0010: Updated the test structure in [9e3dc30](https://github.com/Twingate/kubernetes-operator/pull/728/commits/9e3dc30db69ad49da5ddefc4a2b40464a4258556)

ekampf (approved):

Issue #724, opened 2025-07-17
Twingate related annotations added to every kind=Service in cluster


joshuabaird:
**What happened?**
We have noticed that every kind=_Service_ in clusters that have this operator running have two Twingate-related annotations added to them:

```
    twingate.com/kopf-managed: "yes"
    twingate.com/last-handled-configuration: |
      {"spec":{...}}}
```

Is this to be expected? 

Perhaps related to [this](https://github.com/Twingate/kubernetes-operator/blob/393d410fd3b7f828c26979cc001446725e141d27/app/handlers/handlers_services.py#L216-L217) code?

**Environment**

* Twingate Kubernetes Operator version:

`0.24.0`

* Kubernetes version information:

`1.33`

* Kubernetes cluster kind:

EKS

ekampf:
@joshuabaird this is expected.
The operator has to track _all services_ to know when the `resource.twingate.com` annotation is added or removed so it can then create\delete the related TwingateResource object. It adds this annotation to be able to track this state....

Is this an issue for you?

PR #760, opened 2025-08-23
fix: operator now does not add annotations to unrelated services

Lepird:

- Issue: [724](https://github.com/Twingate/kubernetes-operator/issues/724)

## Changes
updated the service `on.update` handler to properly filter out unrelated services

## Notes
Added This small change to reduce noise on on unrelated/untagged services.

Also the operator was causing extra load on the k8s api server to query available twingate resources for every change on every service on the cluster with this update this is no longer the case

ref [change handlers](https://kopf.readthedocs.io/en/stable/filters/#change-filters)

I have used the Tuple notations to be able to escape the dots in the annotation key which is undocumented but fully supported, ref: [implementation](https://github.com/nolar/kopf/blob/ca3e0d0e2d3147c6728f3f01c4c7d2524f654b36/kopf/_cogs/structs/dicts.py#L23)


app/handlers/handlers_services.py
```diff
@@ -213,8 +213,9 @@ def twingate_service_create(body, spec, namespace, meta, logger, reason, **_):
         )
 
 
-@kopf.on.update("service", annotations={"resource.twingate.com": "false"})
-@kopf.on.update("service", annotations={"resource.twingate.com": kopf.ABSENT})
+#Use Tuple for the field to properly escape dots in the annotation key
```
clement0010, line 216: Lets fix the comment `# Use Tuple for the field to properly escape dots in the annotation key.`

clement0010 (approved):
Fix & merge!

app/handlers/handlers_services.py
Lepird: I have updated the comment

minhtule (approved):
LGTM! 👍 

> I have used the Tuple notations to be able to escape the dots in the annotation key which is undocumented but fully supported, , ref: [implementation](https://github.com/nolar/kopf/blob/ca3e0d0e2d3147c6728f3f01c4c7d2524f654b36/kopf/_cogs/structs/dicts.py#L23)

The `field` argument type is documented to be `None | str | tuple[str, ...] | list[str]` [here](https://kopf.readthedocs.io/en/stable/packages/kopf.on/#kopf.on.update) so I think it should be well supported.

PR #761, opened 2025-08-25
fix: Trim `ResourceProxy.certificate_authority_cert` empty whitespace

clement0010:
## Changes
- The base64-decoded value of  `certificate_authority_cert` in Resource CRD can have leading and trailing whitespace. As a result, the diffing check [here](https://github.com/Twingate/kubernetes-operator/blob/b9ef473b086c1827db2af119370101923a0fd851/app/api/client_resources.py#L220) would evaluate to `False`


app/api/tests/factories.py
```diff
@@ -17,6 +17,8 @@
...
...GpVdnpMeE0xc1d6ZWJ0NEl0M2c4ODZNa1M0bDB3WmdhWUh4bWNtQ2RsSnZ5UHFWOHR4Z1FaWUJZM2pUN0VqZ1BGb3g0a0xNVktBK2pBemY5c0hUaDd6UW5PZ1JFMzJyaGoyTlVBSzNoQmJIdjFhT2VVbGh4U0xEbGU3WDZsWEd4eEhDdkEzbDFOcG1vNUExT1poTUJGQWdNQkFBR2pBakFBTUEwR0NTcUdTSWIzRFFFQkN3VUFBNElDQVFDVUlvcDJUU1FKenNSaGd3T0dZa2JwQWJsU2prTlE1VEJaZnJyWm9GWU9NQTBqaTYycWxXRDNDNU9VYVdRYkJydkcvOEx2Q09YbTRtUG1wMWUwSmVsaTZEWkJJbjJVbzduZTI5VitpdHZnQi9kdTYrcGtySXIwZWdBYmtKZmtTK2YzbFFqZXBqRmFraVFxSzNZTEp0WEpVckt2d2pXa2RnVG1XcjhTMVA5TFg0ZkU0UmxyOWkrcGc2TlZzcFNEZXptREhnN2piZ2NxMXRLOGczcmFEcEFNNExreUdKSENTRTB0V21ORHc2UUtSYi9ldjZmQmR6MVVWVFhhV1pvQTIycldjZk1IMzVZd2NDUDVvWHBpa0lTaStKbUc2SG9qQnM0bGpwYlpGWWNSUnU5UC9pMG12cGRKUXRQUlJudk5DNXY1RXdQdWt0RTFXaTZxa3A2OE43aitRTGw4anlhWExuNkdIRTZDamdnRTRZQjh2ZXFjZUxhRERZdXR4UmpUNzdMaEVTeFdONlhSQnpoTWNPckhGcE5KUUkxVmxhbEFCVzJZanBKSVB2bytpV2xBWlp4MjBrMitHRkpWTml3ZTBYemR5cWwxZUdNeENrS3BkNXdCZXpKZUJVdXJNUSt0cWQrMWRHMTBmRUJMM2dpa0JHbFpMU1d1czNwRnhTaXdZemhTU29BcUs5ekY3VCtBNjc0cDdFUUI5Zms4VjRadFI2T28yMFI0TldPWDRWcnFzekZjWWFOSnJwS3VCOEZhRHZVcWNFMmFRK3ZrWVhmaTBGYWRMRm1jN1dlTWVQSU12a2ZpbnI5cUVnWWMreXE1WGEwV0hiM1hlKzF5N2wwVEt5dUtIZEJnSEdKVUJBYkFFeWF1NFVjSUJqbjFnWDJZTlEvTjFUUnZxSWJrY1E9PQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0t"
```
ekampf, line 18: do we need this const in code?   can this just be `VALID_CA_CERT_1 = base64.b64_encode(...)` ?

clement0010: Yes we could technically use `BASE64_OF_VALID_CA_CERT = base64.b64_encode(...)`. IMO, I think it's nicer to assert the exact base64 string, because we want test to fail if somehow `b64_encode()` changes its implementation, e.g. add trailing padding `\n`, we want to be able to catch these changes. If we rely on `b64encode()` during the test, then the test only verifies that we call `b64encode()`.

minhtule (approved):
LGTM 👍

PR #789, opened 2025-09-16
fix: update CRD schema to set properties to null for oneOf conditions to ensure proper validation

### LINKED ISSUE #787: Rendered manifests CRD issue
**What happened?**

After deploying tg operator with rendered manifests (we use GitOps & ArgoCD) everything works as intended but the CRD details change from manifests that were published (using `helm template` content) compared to actual state.  This causes a diff for desired state to current state and thus ArgoCD will perpetually show it out of sync.  Screenshot included below.  It is not a good idea to ignore differences, especially for CRD contents, so can you please adjust the CRD in the rendered manifests (`helm template` content) to match.  In this case it looks like the `properties` field needs to be added.

**Did you expect to see something different?**

Yes, here's the content snippet that I think would be correct:
```spec:
  versions:
    - additionalPrinterColumns:
      name: v1beta
      schema:
        openAPIV3Schema:
          properties:
            spec:
              oneOf:
                - properties: null
                  required:
                    - resourceRef
                    - principalId
                - properties: null
                  required:
                    - resourceRef
                    - principalExternalRef
                - properties: null
                  required:
                    - resourceRef
                    - groupRef
```

**How to reproduce it (as minimally and precisely as possible)**:

**Environment**

    `twingate/kubernetes-operator:0.26.2`
    

* Kubernetes version information:

    `Server Version: v1.33.2`
    

* Kubernetes cluster kind:

    AKS

* Manifests:

Helm Chart details:
```
apiVersion: builtin
kind: HelmChartInflationGenerator
metadata:
  name: twingate
releaseName: twingate
namespace: twingate
repo: oci://ghcr.io/twingate/helmcharts
name: twingate-operator
includeCRDs: true
valuesFile: ../../../app-templates/twingate/values-base.yaml
additionalValuesFiles:
  - ./values-generated.yaml
  - ./values-custom.yaml
version: "0.26.2"
```
Note: the values yaml content is not relevant as the CRD's do not change.

* Twingate Kubernetes Operator Logs:

Not Applicable

Image

ekampf:

Fixes #787

### DIFF
--- a/deploy/twingate-operator/crds/twingate.com.twingateresourceaccesses.yaml
+++ b/deploy/twingate-operator/crds/twingate.com.twingateresourceaccesses.yaml
@@ -18,11 +18,11 @@ spec:
               type: object
               description: "TwingateResourceAccessSpec defines the desired state of TwingateResourceAccess"
               oneOf:
-                - properties:
+                - properties: null
                   required: ["resourceRef", "principalId"]
-                - properties:
+                - properties: null
                   required: ["resourceRef", "principalExternalRef"]
-                - properties:
+                - properties: null
                   required: ["resourceRef", "groupRef"]
               properties:
                 principalId:


minhtule (approved):
LGTM 👍

PR #807, opened 2025-10-09
feat: Support `certificateAuthorityCertSecretRef` on `TwingateResource` CRD

clement0010:
## Changes
- Add `certificateAuthorityCertSecretRef` on `TwingateResource` CRD. This enables Kubernetes Resource CRD to reference K8s secret object to update/reconcile the certificate authority cert value on Twingate via Public API.
- Extract `k8s_get_secret` to `utils_k8s`
- Move `get_ca_cert` to `ResourceProxy`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -195,7 +195,7 @@ jobs:
           poetry version -s
 
       - name: Login to Docker Hub
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -109,7 +109,7 @@ jobs:
           installer-parallel: true
 
       - name: Login to Docker Hub
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/.github/workflows/vulnerability-monitor.yaml
+++ b/.github/workflows/vulnerability-monitor.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Login to Docker Hub
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}


app/crds.py
```diff
@@ -122,8 +125,19 @@ class ResourceProxy(BaseModel):
...
     certificate_authority_cert: Annotated[
-        Base64Str, AfterValidator(lambda v: v.strip())
-    ]
+        Base64Str | None, AfterValidator(lambda v: v.strip() if v is not None else None)
+    ] = None
+    certificate_authority_cert_secret_ref: _KubernetesObjectRef | None = None
+
+    def get_certificate_authority_cert(self) -> str | None:
+        if secret_ref := self.certificate_authority_cert_secret_ref:
+            secret = k8s_get_secret(secret_ref.namespace, secret_ref.name)
```
ekampf: Wby no walrus :(
```
f secret_ref := self.certificate_authority_cert_secret_ref:
  if secret := k8s_get_secret(secret_ref.namespace, secret_ref.name):
    return base64.b64decode(get_ca_cert(secret)).decode()
  
  return None

```

app/utils_k8s.py
```diff
@@ -40,3 +45,32 @@ def k8s_read_namespaced_deployment(
         if ex.status == 404:
             return None
         raise
+
+
+def k8s_get_secret(namespace: str, name: str) -> kubernetes.client.V1Secret | None:
```
ekampf: Other methods in this module also get the `kapi: kubernetes.client.CoreV1Api | None = None` kwarg and only instanciate a new `CoreV1API` if not given one.
Lets keep the same behavior here?

app/utils_k8s.py
```diff
@@ -40,3 +45,32 @@ def k8s_read_namespaced_deployment(
...
+            name=name, namespace=namespace
+        )
+    except kubernetes.client.exceptions.ApiException as ex:
+        if ex.status == 404:
+            return None
+
+        raise
+
+
+def get_ca_cert(secret: kubernetes.client.V1Secret) -> str:
```
ekampf: I wonder if this should be here because its not purely a "k8s utility" thats general - its very specific to the CA secret of `ResourceProxy` - maybe it should be there?

app/handlers/handlers_services.py
```diff
@@ -140,7 +107,11 @@ def service_to_twingate_resource(service_body: Body, namespace: str) -> dict:
                     if spec["type"] == ServiceType.LOAD_BALANCER
                     else f"{service_name}.{namespace}.svc.cluster.local"
                 ),
-                "certificateAuthorityCert": get_ca_cert(tls_secret),
+                "certificateAuthorityCert": base64.b64encode(
+                    ResourceProxy.read_certificate_authority_cert_from_secret(
+                        secret
+                    ).encode()
+                ).decode(),
```
clement0010: This is conversion is temporary, it'll be removed in this [PR](https://github.com/Twingate/kubernetes-operator/pull/808/files#diff-8703dce0ccbc9d51ff51eeaab1da0346af9682f282ff0ebd6f275d96e84007b9L110-L114)

PR #820, opened 2025-10-24
feat: Add PodDisruptionBudget twingate operator

twingate-blee:
## Summary

This PR adds configurable PodDisruptionBudget support to the twingate-operator Helm chart.

## Changes

- Created new PDB template (`pdb.yaml`) with values-driven configuration
- Added PDB configuration to `values.yaml` (disabled by default for backward compatibility)
- Supports both `minAvailable` and `maxUnavailable` configurations
- Added comprehensive test suite (`pdb_test.yaml`) with 5 test cases covering:
  - Disabled PDB behavior
  - minAvailable configuration
  - maxUnavailable configuration
  - Metadata validation
  - Selector labels validation

## Configuration

```yaml
pdb:
  create: false        # Set to true to enable PDB
  minAvailable: 1      # Minimum number of pods that must remain available
  maxUnavailable: ""   # Maximum number/percentage of pods that can be unavailable
```

## Test Plan

- Unit tests added in `pdb_test.yaml`
- Helm chart linting passes
- Template rendering verified

### DIFF
--- a/deploy/twingate-operator/values.yaml
+++ b/deploy/twingate-operator/values.yaml
@@ -130,7 +130,7 @@ extraEnvVars: []
 pdb:
   create: false
   # minAvailable: 1
-  # maxUnavailable:
+  # maxUnavailable: ""
 
 rbac:
   ## Create ClusterRoles that extend the existing view, edit and admin ClusterRoles to interact with Twingate's CRDs


deploy/twingate-operator/values.yaml
```diff
@@ -127,6 +127,11 @@ priorityClassName: ""
 extraEnvVars: []
 
 
+pdb:
+  create: false
+  # minAvailable: 1
+  # maxUnavailable:
```
minhtule: Missing value?

deploy/twingate-operator/tests/pdb_test.yaml
```diff
@@ -0,0 +1,80 @@
...
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: should create PDB with minAvailable when pdb.create is true
+    set:
+      pdb:
+        create: true
+        minAvailable: 1
+        maxUnavailable: ""
```
minhtule: Is this empty string intentional? or can we just omit it?

twingate-blee: Good catch. I omitted it

deploy/twingate-operator/values.yaml
minhtule: Also I just noticed other string properties like `nameOverride: ""` uses empty string as the template so using `""` would be more consistent with the rest of the file.

PR #822, opened 2025-10-27
fix: Fix the reconciliation logic for Kubernetes resource

minhtule:
## Changes

- Fix the reconciliation for Kubernetes resource
  + Add port `443` explicitly to Gateway address in `TwingateResource` CRD.
    - Gateway address includes the port number in the backend so the CRD should also include the port number toto avoid difference between CRD data and what is returned in public API. Right now the CRD stores `gateway.svc.default.local.cluster.local` while the API returns `gateway.svc.default.local.cluster.local:443`, which results in unnecessary update mutations.
  + When comparing the certificates, convert them to [`cryptography.x509.Certificate`](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate)
    - Right now we compare certificates in PEM format but the same certificate can be formatted differently in PEM.

- Add integration test to make sure we don't make unnecessary update mutation for Kubernetes resources

### DIFF
--- a/tests_integration/conftest.py
+++ b/tests_integration/conftest.py
@@ -66,8 +66,8 @@ def run_kopf(kopf_runner_args, kopf_settings):
     @contextmanager
     def inner(
         *,
-        enable_connector_reconciler=True,
-        enable_group_reconciler=True,
+        enable_connector_reconciler=False,
+        enable_group_reconciler=False,
         enable_resource_reconciler=False,
         cleanup=True,
     ):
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -67,7 +67,7 @@ test-cov: ##@tests Runs all tests with coverage
 
 .PHONY: test-int
 test-int: ##@tests Runs integration tests
-	poetry run pytest  -m "integration" -vv -l -x
+	poetry run pytest ./tests_integration -m "integration" -vv -l -x
```
minhtule: This is to prevent pytest from loading the `apps.handers` module when collecting test files. That would prevent us from patching the envvars for kopf timers like [this one](https://github.com/Twingate/kubernetes-operator/blob/37873f05e5997d3824e6ea4acdeb44c5f4202627/app/handlers/handlers_resource.py#L94-L96).

clement0010 (approved):
LGTM 👍🏻

ekampf (approved):

PR #826, opened 2025-11-05
fix: Don't set `TWINGATE_DEFAULT_RESOURCE_TAGS` envvar when not provided

clement0010:
## Changes
- Check if `defaultResourceTags` value is provided when setting `TWINGATE_DEFAULT_RESOURCE_TAGS` envvar


deploy/twingate-operator/templates/deployment.yaml
```diff
@@ -68,8 +68,10 @@ spec:
             value: {{ required "Network name required" .Values.twingateOperator.network }}
           - name: TWINGATE_HOST
             value: {{ .Values.twingateOperator.host | default "twingate.com" }}
+          {{- if .Values.twingateOperator.defaultResourceTags }}
           - name: TWINGATE_DEFAULT_RESOURCE_TAGS
             value: {{ .Values.twingateOperator.defaultResourceTags | default dict | toJson | quote }}
```
minhtule: do we still need `default dict` here?

clement0010: Good catch. Nope, we don't need it

PR #859, opened 2025-12-16
chore: Pin helm-unittest version to v1.0.3

minhtule:
## Changes

Pin helm-unittest to v1.0.3 until we upgrade to Helm v4 because helm-unittest is causing [installation issue](https://github.com/helm-unittest/helm-unittest/issues/790)


clement0010 (approved):
LGTM 👍🏻

Issue #893, opened 2026-02-19
TLS certificate regeneration during GitOps reconciliation breaks operator-to-control-plane sync + gateway cert_reloader has fragile fsnotify handling


jescarri:
**What happened?**

When using ArgoCD (or any GitOps tool) to manage the Twingate Kubernetes Access Gateway Helm chart, the self-signed TLS secret (twingate-operator-kubernetes-access-gateway-tls) is regenerated on every reconciliation cycle. After the certificate is regenerated, the operator fails to re-register the new CA certificate with the Twingate control plane, leaving the entire Kubernetes Access Gateway in an unusable state until the operator pod is manually restarted.

There are three related issues here:

---

### Issue 1: Helm `lookup` does not work with ArgoCD by default (documentation gap)

The gateway Helm chart uses [`lookup`](https://github.com/Twingate/kubernetes-access-gateway/blob/v0.11.1/deploy/gateway/templates/tls-secret.yaml#L4) to check for an existing TLS secret and reuse its cert/key data when the alternative names haven't changed:

```yaml
{{- $existingSecret := lookup "v1" "Secret" .Release.Namespace $secretName }}
```

However, Helm `lookup` **returns empty during ArgoCD's default client-side dry-run** rendering. This means ArgoCD always sees the TLS secret as "new", regenerates the self-signed certificate/key/CA on every sync, and applies it — even when nothing has actually changed.

The workaround is to set `source.helm.serverDryRun: true` (or use server-side rendering) in the ArgoCD Application manifest, but **this is not documented anywhere** in the chart's README or ArgoCD integration docs.

**Ask:** Please document the required ArgoCD configuration for proper `lookup` behavior in the chart README and/or an ArgoCD-specific installation guide.

### Issue 2: Operator does not sync new CA to the Twingate control plane after TLS secret change (bug)

This is the **core issue**. The sequence of events:

1. ArgoCD syncs → TLS secret is regenerated with a **new** self-signed cert, key, and CA.
2. The gateway pod detects the volume change and reloads the new cert/key (confirmed by logs — see below).
3. The **operator**, however, does **not** detect that the TLS secret (specifically the CA) has changed.
4. The Twingate control plane still has the **old** CA registered.
5. Any connection routed through the gateway now fails TLS verification because the gateway presents a cert signed by the **new** CA, but the control plane only trusts the **old** CA.
6. Result: `connection reset by peer` errors for all traffic through the gateway.

[...]

PR #909, opened 2026-03-06
feat: Add K8s TLS secret object watch handler

clement0010:

- Issue: https://github.com/Twingate/kubernetes-operator/issues/893

## Changes
- Add TLS secret event handler.
  - Detect changes on the `data["ca.crt"]` field and update Twingate Resource's CA cert value
- Add `twingate_resource_secret_index` to build a mapping of secrets by TwingateResources
  - When a secret is updated, we can lookup the index to see whether it is used by any TwingateResources


app/handlers/handlers_secrets.py
```diff
@@ -0,0 +1,47 @@
...
+        resource = client.resource_update(
+            resource_type=crd.type,
+            **crd.to_graphql_arguments(labels=labels),
+        )
+        logger.info("Got resource %s", resource)
+    except GraphQLMutationError:
+        logger.exception("Failed to update resource %s after secret changed", crd.id)
+
+
+@kopf.on.update("", "v1", "secrets", field=("data", "ca.crt"))  # type: ignore[arg-type]
```
clement0010: If we want to avoid `kopf` from annotating other non-related TLS secret object, we could use `@kopf.on.event()`. The downside is that we need to do the diffing ourselves.

app/handlers/handlers_secrets.py
```diff
@@ -0,0 +1,95 @@
...
+
+from app.api import TwingateAPIClient
+from app.api.client_resources import NetworkResource
+from app.api.exceptions import GraphQLMutationError
+from app.crds import ResourceSpec
+from app.handlers.handlers_services import k8s_get_twingate_resource
+from app.utils import validate_pem_x509_certificate
+
+
+@kopf.on.event("", "v1", "secrets", field="type", value="kubernetes.io/tls")  # type: ignore[arg-type]
```
minhtule: I think we should remove this type filter. While the Gateway Helm chart would create a `kubernetes.io/tls` secret by default, the user could provide any secret type as long as it contains the CA cert in the `data[ca.cert]` field.

clement0010: >  the user could provide any secret type as long as it contains the CA cert in the data[ca.cert] field.

`@kopf.on.event("", "v1", "secrets", field=("data", "ca.crt"))` should watch for what you just described. Updated in [3bd86e6](https://github.com/Twingate/kubernetes-operator/pull/909/commits/3bd86e6b1076f6e686e1fb8546970b92eb6f3ecb)

tests_integration/test_resource_flows.py
```diff
@@ -271,6 +272,85 @@ def test_resource_created_before_operator_runs(run_kopf, unique_resource_name):
...
+"""
```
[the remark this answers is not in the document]
clement0010: Cleaned up in [976676a](https://github.com/Twingate/kubernetes-operator/pull/909/commits/976676a312e753fbdda875ff857bccc6d2dacf52) to reduce the TLS secret setup duplication

PR #964, opened 2026-04-20
chore: Fix flaky connector flow integration test

clement0010:

- Issue: https://github.com/Twingate/kubernetes-operator/issues/940

## Changes
- Update tests to poll for a specific update message returned by the `twingate_connector_update` handler to ensure the connector deployment is updated.
- Update `twingate_connector_update` to write a new `"Update completed"` success message instead of keeping the old `"No update required"` message.

## Notes
The test is flaky because `time.sleep` is too short for the Connector deployment to reconcile. We should replace `time.sleep` with a better wait method. 

The reason we need to update the message is because: after the initial `kubectl_create`, the Kopf update handler may fire when `twingate_connector_create` patches the spec with the connector ID. That first update returns `"No update required"`. When we do `kubectl_patch`, the next update returns but the old `"No update required"` message is still there. The existing `kubectl_wait_object_handler_success` couldn't differentiate between these two handler invocations.

### DIFF
--- a/app/handlers/handlers_connectors.py
+++ b/app/handlers/handlers_connectors.py
@@ -276,7 +276,7 @@ def twingate_connector_update(body, memo, logger, new, diff, status, namespace, 
 
     create_or_replace_deployment(body, namespace, crd, memo.twingate_settings.full_url)
 
-    return success(twingate_id=updated_connector.id)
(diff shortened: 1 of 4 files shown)


minhtule (approved):
LGTM 👍

Issue #904, opened 2026-03-04
`TwingateResourceAccess` very slow to retry after `ResourceNotFound`


gregbrowndev:
**What happened?**

I had some challenges getting started with the Twingate Operator after finding the service-annotation approach doesn't support inline access config. I migrated to use explicit CRDs for `TwingateResource` and `TwingateResourceAccess`, and allowed ArgoCD to sync. However, I've got into a state where the resources have been applied successfully, but the `TwingateResourceAccess` failed with `ResourceNotFound`. ArgoCD shows the sync was successful. 

```txt
Status:
  twingate_resource_access_change:
    Error:    Resource twingate/carta-ui not found
    Success:  false
  twingate_resource_access_sync:
    Error:    Resource twingate/carta-ui not found
    Success:  false
Events:
  Type     Reason            Age   From  Message
  ----     ------            ----  ----  -------
  Normal   Logging           24m   kopf  Got a TwingateResourceAccess create request: {'principalExternalRef': {'name': 'Everyone', 'type': 'group'}, 'resourceRef': {'name': 'carta-ui', 'namespace': 'twingate'}}
  Normal   Logging           24m   kopf  Handler 'twingate_resource_access_change' succeeded.
  Warning  ResourceNotFound  24m   kopf  Resource twingate/carta-ui not found
  Normal   Logging           24m   kopf  Creation is processed: 1 succeeded; 0 failed.
  Normal   Logging           23m   kopf  Got a TwingateResourceAccess create request: {'principalExternalRef': {'name': 'Everyone', 'type': 'group'}, 'resourceRef': {'name': 'carta-ui', 'namespace': 'twingate'}}
  Warning  ResourceNotFound  23m   kopf  Resource twingate/carta-ui not found
  Normal   Logging           23m   kopf  Timer 'twingate_resource_access_sync' succeeded.
```

**Did you expect to see something different?**

I would expect the operator to retry more quickly, e.g. after seeing a new resource being created. I've waited 10 mins or so and it doesn't look like it will fix itself.

I'm not sure if it would be possible but it would be helpful if the failure could be exposed to ArgoCD so it could show the sync has failed.

**How to reproduce it (as minimally and precisely as possible)**:

- Create a `TwingateResourceAccess` referencing a `TwingateResource` that doesn't exist yet
- Wait to see the `TwingateResourceAccess` fail with `ResourceNotFound`
- Create the `TwingateResource`, the `ResourceNotFound` error is not resolved 

[...]

PR #979, opened 2026-05-05
fix: retry TwingateResourceAccess quickly when referenced TwingateResource is missing (#904)

ekampf:

- Issue: #904

## Changes

- Raise `kopf.TemporaryError` (delay=30s) instead of returning a failure dict when a `TwingateResourceAccess` references a `TwingateResource` that does not exist yet — Kopf will now retry the handler within ~30s instead of waiting up to 10 hours for the reconciliation timer.
- Preserve the existing `kopf.warn(reason="ResourceNotFound", ...)` so the K8s Event remains visible to `kubectl describe` and to ArgoCD.
- Update `test_create_invalid_ref` to assert the new raise behavior while still verifying the warning event is emitted.

### Steps to reproduce (from the issue)

1. Apply a `TwingateResourceAccess` whose `resourceRef` points to a `TwingateResource` that hasn't been created yet (common with ArgoCD's arbitrary apply order).
2. Observe `Warning ResourceNotFound` in `kubectl describe`.
3. Apply the `TwingateResource`.
4. **Before this fix:** the access stays broken until the 10-hour `twingate_resource_access_sync` timer fires.
5. **After this fix:** the access reconciles within ~30 seconds.

### Root cause

In `app/handlers/handlers_resource_access.py`, the "resource not found" branch was returning a failure dict, which Kopf treats as a completed handler — no retry. The adjacent "resource exists but has no id yet" branch was already correctly raising `kopf.TemporaryError`. The fix aligns the two branches.

### Alternatives considered

- **Kopf indexers / event-driven reconcile of dependents** (watching `TwingateResource` create events to trigger immediate reconciliation of dependent `TwingateResourceAccess` resources). More responsive than polling, but a non-trivial architectural change. Deferred — 30-second retry resolves the user-visible pain.
- **Lowering the 10-hour `twingate_resource_access_sync` interval.** The long interval is intentional (see comment at `handlers_resource_access.py:115-117`) — tenants with many access CRDs can hit rate limits. Not changed.
- **Surfacing the failure as a CRD status condition for ArgoCD.** ArgoCD already sees the K8s Event from `kopf.warn` plus the existing status payload. Larger change with backwards-compat considerations — not needed for this fix.

### DIFF
--- a/app/handlers/handlers_resource_access.py
+++ b/app/handlers/handlers_resource_access.py
@@ -74,7 +74,7 @@ def twingate_resource_access_change(body, spec, memo, logger, patch, status, **k
     if not resource_crd:
         err = f"Resource {access_crd.resource_ref_fullname} not found"
         kopf.warn(body, reason="ResourceNotFound", message=err)
-        return {"success": False, "error": err}
+        raise kopf.TemporaryError(err, delay=30)
 
     if not resource_crd.spec.id:
         raise kopf.TemporaryError("Resource not yet created, retrying...", delay=15)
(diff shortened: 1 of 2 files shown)


minhtule (approved):
LGTM 👍

PR #994, opened 2026-05-12
feat: support full AccessInput in resource access API

ekampf:
## Changes

- Extend `AccessInput` in `app/api/client_resources_access.py` with `expiresAt`, `accessPolicy`, and `approvalMode` so all five `AccessInput` GraphQL fields (per https://www.twingate.com/docs/api) are supported. `resource_access_add` accepts the new values as optional kwargs.
- Add `AccessMode`, `AccessApprovalMode`, and `AccessPolicyInput` to `app/crds.py` (shared with the client to avoid circular imports — same pattern as `ConnectorSpec`/`ResourceSpec`/`GroupSpec`).
- Extend `ResourceAccessSpec` and the `twingateresourceaccesses` CRD YAML with `expiresAt` (RFC3339), `accessPolicy` (`{mode, durationSeconds}`), and `approvalMode` (`MANUAL` | `AUTOMATIC`).
- Handler forwards the new fields straight through to `client.resource_access_add`.
- Switch the `AccessInput` GraphQL payload to `model_dump(by_alias=True, mode="json")` so `datetime` and enum values serialize correctly.

### CRD schema validations
- `accessPolicy.mode` is an `enum` of `MANUAL` / `AUTO_LOCK` / `ACCESS_REQUEST`; required when `accessPolicy` is set.
- CEL rule: `AUTO_LOCK` requires `durationSeconds`; `MANUAL` forbids it. Caught at `kubectl apply` time.
- `durationSeconds` bounded between `3600` (1 hour) and `31536000` (365 days).
- `approvalMode` is an `enum` of `MANUAL` / `AUTOMATIC`.

### Review feedback addressed
- `StrEnum` used for all access-related enums.
- `CLAUDE.md` updated with a "CRD Schema Changes" rule: any new enum/min/max/pattern/required/CEL/oneOf must be paired with a `tests_integration/test_crds_*.py` test (pydantic doesn't enforce OpenAPI/CEL, only the apiserver does).
- Test naming cleanup: `test_resource_access_add_success` renamed to `test_resource_access_add_with_security_policy_id`; duplicate `test_resource_access_add_success_without_policy` removed (covered by the new `test_resource_access_add_omitted_fields_serialize_as_null`).

## Notes

- The Twingate docs state that `securityPolicyId`, `expiresAt`, `accessPolicy`, and `approvalMode` "should be null when the principal is a Service Account." We do **not** enforce this client-side — the Twingate API will reject invalid combinations and the existing `GraphQLMutationError` path surfaces them. The new CRD field descriptions call this out for kubectl users.
- The "omit field in CRD spec ⇒ explicit `null` in GraphQL payload ⇒ clear server-side value" semantic that `securityPolicyId` already follows is preserved for the three new fields (per the Twingate docs).

## Tests

- 4 new client tests (`expires_at`, `access_policy`, `approval_mode`, omitted-fields-serialize-as-null) + the 3 existing add tests updated for the new payload keys.
- 3 new handler tests forwarding `expiresAt` / `accessPolicy` / `approvalMode` to the client.
- 12 new integration tests in `tests_integration/test_crds_resourceaccess.py` covering every new CRD validation (7 negative for CEL / bounds / enum / required, 5 positive happy paths).
- `make typecheck` clean; `make test` 258 passed locally.

## Test plan

- [ ] `make typecheck`
- [ ] `make test`
- [ ] `make test-helm`
- [ ] `make test-int` (covers all 12 new CRD validation cases against a real apiserver)
- [ ] Manual: apply a `TwingateResourceAccess` with `expiresAt`, `accessPolicy.mode: AUTO_LOCK` + `durationSeconds`, and `approvalMode: MANUAL`; verify the access edge in the Twingate admin console reflects all three.


app/crds.py
```diff
@@ -262,6 +262,24 @@ class PrincipalTypeEnum(str, Enum):
     ServiceAccount = "serviceAccount"
 
 
+class AccessMode(str, Enum):
```
clement0010: nit: It'd be nicer to use `StrEnum`

app/crds.py
```diff
@@ -262,6 +262,24 @@ class PrincipalTypeEnum(str, Enum):
...
+    MANUAL = "MANUAL"
+    AUTO_LOCK = "AUTO_LOCK"
+    ACCESS_REQUEST = "ACCESS_REQUEST"
+
+
+class AccessApprovalMode(str, Enum):
```
clement0010: ditto

deploy/twingate-operator/crds/twingate.com.twingateresourceaccesses.yaml
```diff
@@ -70,6 +70,35 @@ spec:
...
+                      type: string
+                      enum:
+                        - MANUAL
+                        - AUTO_LOCK
+                        - ACCESS_REQUEST
+                      description: "Access mode."
+                    durationSeconds:
+                      type: integer
+                      nullable: true
+                      description: "Duration of the access policy in seconds. Used with AUTO_LOCK mode."
```
clement0010: It can also be used with `ACCESS_REQUEST` mode
```suggestion
                      description: "Duration of the access policy in seconds. Used with AUTO_LOCK and ACCESS_REQUEST mode."
```

deploy/twingate-operator/crds/twingate.com.twingateresourceaccesses.yaml
```diff
@@ -70,6 +70,35 @@ spec:
...
+                  type: string
+                  format: date-time
+                  nullable: true
+                  description: "RFC3339 datetime when this access expires. Omitting the field or setting it to null clears any existing expiration. Must be null when the principal is a ServiceAccount."
+                accessPolicy:
+                  type: object
+                  nullable: true
+                  required: ["mode"]
+                  description: "Access policy applied to this access edge. Omitting the field or setting it to null clears any existing policy. Must be null when the principal is a ServiceAccount."
+                  properties:
```
clement0010, line 88: I think we can enforce these validation rules:
```yaml
x-kubernetes-validations:
  - rule: "self.mode != 'AUTO_LOCK' || has(self.durationSeconds)"
    message: "durationSeconds is required when mode is AUTO_LOCK"
  - rule: "self.mode != 'MANUAL' || !has(self.durationSeconds)"
    message: "durationSeconds must not be set when mode is MANUAL"
```

so we can fail at `kubectl apply` instead of failing at the API call

deploy/twingate-operator/crds/twingate.com.twingateresourceaccesses.yaml
```diff
@@ -70,6 +70,35 @@ spec:
...
+                    mode:
+                      type: string
+                      enum:
+                        - MANUAL
+                        - AUTO_LOCK
+                        - ACCESS_REQUEST
+                      description: "Access mode."
+                    durationSeconds:
```
clement0010, line 96: we can add minimum: 3600 (1hour) and maximum: 31536000 (365 days) validation

app/api/tests/test_client_resource_access.py
```diff
@@ -95,6 +118,155 @@ def test_resource_access_add_failure(self, test_url, api_client, mocked_response
...
+            ],
+        )
+
+        assert api_client.resource_access_add(
+            resource_id=resource_id,
+            principal_id=principal_id,
+            approval_mode=AccessApprovalMode.AUTOMATIC,
+        )
+
+    def test_resource_access_add_omitted_fields_serialize_as_null(
```
clement0010, line 202: This test is identical to `test_resource_access_add_success_without_policy`, I think we can remove `test_resource_access_add_success_without_policy`

app/api/tests/test_client_resource_access.py
clement0010: nit: lets rename this to `test_resource_access_add_with_security_policy_id` for consistency with other test case

app/api/client_resources_access.py
```diff
@@ -41,12 +47,19 @@ def resource_access_add(
...
         access = AccessInput(
-            principal_id=principal_id, security_policy_id=security_policy_id
+            principal_id=principal_id,
+            security_policy_id=security_policy_id,
+            expires_at=expires_at,
+            access_policy=access_policy,
+            approval_mode=approval_mode,
         )
-        access_list = [access.model_dump(by_alias=True)]
+        access_list = [access.model_dump(by_alias=True, mode="json")]
```
minhtule, line 62: why do we need to specify the `mode` now? 🤔

minhtule (approved):
Fix & merge!

app/api/client_resources_access.py
ekampf: The default `model_dump` returns Python-native values — `datetime` stays as `datetime`, `Enum` members stay as enum instances. The old payload was all `str | None` so that worked. The new fields are:

- `expires_at: datetime | None`
- `access_policy.mode: AccessMode` (enum)
- `approval_mode: AccessApprovalMode` (enum)

Without `mode="json"`, gql's JSON encoder hits the raw `datetime`/`Enum` objects and raises `TypeError`. `mode="json"` makes pydantic emit ISO 8601 strings for datetimes and `.value` strings for enums, which is what the Twingate API expects.

minhtule (approved):
LGTM 👍

clement0010 (approved):
LGTM 👍🏻

clement0010:
@ekampf I think we need to add a GH Issue in the PR description

Issue #997, opened 2026-05-14
Resolve shard URL at Operator startup


clement0010:
## Problem

The Twingate operator currently sends all API requests to the base host (e.g., `twingate.com`), which results in a 308 redirect on every request to the correct shard (e.g., `us1.twingate.com`). This adds unnecessary latency to every API call.

## Solution

At operator startup, perform a HEAD request to resolve the correct sharded host via the 308 redirect, then cache the resolved host for all subsequent API calls. This ensures all GraphQL requests go directly to the correct shard.

## Acceptance Criteria

- On startup, operator resolves the sharded host from the 308 redirect
- Shard resolution happens before remote network name lookup
- Retries with exponential backoff on transient failures
- Falls back to original host if resolution fails
- Sends `User-Agent: Twingate-Operator/<version>` header
- Unit tests cover: successful redirect, non-308 status, connection errors, retry-then-succeed, missing/bad Location header, already-sharded host, call ordering

## Note
Shard URL is not supported in production yet. This is preparation work, so the operator is ready when sharding is enabled in production.

PR #998, opened 2026-05-14
feat: Resolve shard URL at Operator startup

clement0010:
## Related Tickets

Issue: #997

## Changes

- Add `resolve_shard_host()` in `app/settings.py` that performs a `HEAD` request at startup to detect 308 redirects and cache the resolved sharded host (e.g. `us1.twingate.com`)
  - Retries up to 5 times with exponential backoff via tenacity; falls back to the original host on failure 
  - Include `User-Agent: Twingate-Operator/<version>` header on each shard resolution request
- Shard resolution should run before the first remote network name lookup in `TwingateOperatorSettings.__init__()`

## Notes
Sharded host is not supported in production yet. This is preparation work, so the operator is ready when sharding is enabled in production.


app/settings.py
```diff
@@ -3,16 +3,58 @@
...
+        logger.info("Resolved shard host %r -> %r", host, sharded_host)
+        return sharded_host
+
+    return host
+
+
+def get_host(network: str, host: str) -> str:
+    try:
+        return _resolve_shard_host(network, host)
+    except:
```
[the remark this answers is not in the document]
clement0010: We want to fallback to the default host when there's any exception

app/settings.py
```diff
@@ -3,16 +3,63 @@
...
+        timeout=10,
+        headers={"User-Agent": f"Twingate-Operator/{get_version()}"},
+    )
+    location = response.headers.get("location")
+    if response.status_code != 308 or not location:
+        return host
+
+    hostname = urlparse(location).hostname or ""
+    prefix = f"{network}."
+    if hostname.startswith(prefix) and hostname.endswith(f".{host}"):
```
ekampf: not sure why we need the `endwith` check but isnt the return here should just be `hostname[len(prefix):]` ?  simpler than lstrip

app/settings.py
```diff
@@ -3,16 +3,63 @@
...
+    stop=stop_after_attempt(5),
+    retry=retry_if_exception_type(requests.RequestException),
+    reraise=True,
+    wait=wait_exponential(multiplier=0.1, max=10),
+)
+def _resolve_shard_host(network: str, host: str) -> str:
+    url = f"https://{network}.{host}/api/graphql/"
+    response = requests.head(
+        url,
+        timeout=10,
```
minhtule: - 10s + 5 retries would result in ~50s total, probably a bit too long
- The backoff is also very small (sub second) i.e. retry happens quickly and the exponential backoff doesn't mean much.

I would use something like this.
```python
@retry(
    stop=stop_after_attempt(3),
    retry=retry_if_exception_type(requests.RequestException),
    reraise=True,
    wait=wait_exponential(multiplier=1, max=10),
)
def _resolve_shard_host(network: str, host: str) -> str:
    url = f"https://{network}.{host}/api/graphql/"
    response = requests.head(
        url,
        timeout=3,
    )
```

app/settings.py
```diff
@@ -3,16 +3,63 @@
...
+        headers={"User-Agent": f"Twingate-Operator/{get_version()}"},
```
minhtule: I would extract this user agent and also use it in [TwingateRequestsHTTPTransport](https://github.com/Twingate/kubernetes-operator/blob/30700f688c39a9477c44ad7518b3dbff6f40282a/app/api/client.py#L40)

minhtule (approved):
Fix & merge!

app/settings.py
clement0010: Fixed in [23deffe](https://github.com/Twingate/kubernetes-operator/pull/998/commits/23deffe7406aaf3ef8c69f9085736ff41cc4896c)

app/settings.py
clement0010: Fixed in [fd9f692](https://github.com/Twingate/kubernetes-operator/pull/998/commits/fd9f692c4fd65776af3914cba25d00df93ef9fb8)

app/settings.py
```diff
@@ -3,16 +3,63 @@
...
+    return host
+
+
+def get_host(network: str, host: str) -> str:
+    try:
+        resolved_host = _resolve_shard_host(network, host)
+        logger.info("Resolved host %s", resolved_host)
+        return resolved_host
+    except:
+        logger.warning(
```
ekampf: We want to see the exception here. So either `logger.exception` or keep warning but with exc_info true?

ekampf (approved):

PR #1031, opened 2026-06-08
chore: upgrade to Python 3.14

ekampf:
## Related Tickets & Documents

- Issue: Closes #889

## Changes

- `pyproject.toml`: `python = ">=3.14,<3.15"` (single-version, matching the deployed-app pattern)
- `.tool-versions`: `python 3.14.5`
- `.github/workflows/ci.yaml` + `release.yaml`: `PYTHON_VERSION` `3.12` → `3.14`
- `.pre-commit-config.yaml`: pyupgrade `--py311-plus` → `--py314-plus`
- `CLAUDE.md`: stack line → Python 3.14
- `poetry.lock`: regenerated for 3.14; refreshed transitive deps that lacked 3.14 wheels (e.g. `frozenlist` 1.4.0 → 1.8.0, plus multidict/yarl/propcache/etc.)

No `Dockerfile` change needed — its base image is `python:${PYTHON_VERSION}-slim`, driven by the build-arg (CI env / Makefile default).

## Notes

- **Dependency compatibility verified** — all native deps ship 3.14 wheels: pendulum, orjson, pydantic-core (2.41.5+), grpcio (via google-cloud-artifact-registry); cryptography uses `cp311-abi3` (forward-compatible). kopf has no upper Python bound.
- **Verified locally on 3.14.5:** `make check` (ruff, mypy strict, bandit, hadolint) clean; `make test` → 269 passed; `make image-build` builds on `python:3.14-slim` and the resulting image runs 3.14.5 with all native deps importing.
- Independent of #1030 (already merged to main — main already carries kopf 1.44.6 / click 8.4.1, so no conflict).

PR #1030, opened 2026-06-08
chore: unblock kopf updates by running semantic-release via pipx

ekampf:
## Related Tickets & Documents

- Issue: Closes #1029

## Changes

- Remove `python-semantic-release` from the poetry `build` group in `pyproject.toml` (the `[tool.semantic_release]` config stays).
- Run semantic-release via `pipx run --spec python-semantic-release==10.5.3 semantic-release ...` in `scripts/release.sh` and the `release_dev` / `release_dev_helm` jobs in `.github/workflows/ci.yaml` (pipx is pre-installed on GitHub runners).
- Add a pipx pre-flight check to `scripts/release.sh` that exits early with an install hint if pipx is missing.
- Re-resolve `poetry.lock`: drops semantic-release + transitive deps and bumps **kopf 1.38.0 → 1.44.6** and **click 8.1.7 → 8.4.1**.
- Fix two type errors surfaced by kopf 1.44's stricter decorator return types: drop a now-redundant `cast(str, ...)` in `handlers_services.py`, and add a targeted `# type: ignore[attr-defined]` for `twingate_resource_access_change.__name__` (the decorated handler is a plain function at runtime).

## Notes

**Root cause:** Dependabot stopped updating `kopf`. kopf 1.39+ requires `click>=8.2.0`, but `python-semantic-release` (incl. latest 10.5.3) pins `click~=8.1.0`. Because semantic-release was in the poetry build group it participated in lock resolution, capping click at 8.1.7 and making any kopf bump unresolvable — so Dependabot silently opened no PR. No released version combination resolves it; the fix is to stop resolving semantic-release inside the project environment.

Moving it to `pipx run` isolates it from the project venv. Note `pipx run` needs `--spec` here because the package name (`python-semantic-release`) differs from the CLI name (`semantic-release`).

**Verification:** `make check` (ruff, mypy, bandit, hadolint) and `make test` (269 passed) pass locally. `pipx run --spec python-semantic-release==10.5.3 semantic-release --strict version --print` runs and behaves correctly on a feature branch.

PR #1032, opened 2026-06-08
ci: make dependency cache keys job-unique and poetry-aware

ekampf:

- Issue: n/a

## Changes

Improve the `actions/cache` keys in `.github/workflows/ci.yaml`:

- **Include the poetry version** (`env.POETRY_VERSION`) in every cache key, so upgrading poetry no longer reuses a `.venv` built by a different poetry.
- **Make keys unique per job** via `${{ github.job }}`, so `unit-tests` and `e2e-tests` no longer share the same `.venv` key (previously only the first job to finish could save it, and they could cross-contaminate).
- **Fix a latent bug**: the keys interpolated `${{ steps.setup_python.outputs.python-version }}`, but no step defined `id: setup_python` — so it expanded to an empty string and the Python version was effectively absent from the key. Added `id: setup_python` to the `setup-python` steps in `unit-tests` and `e2e-tests`.
- **Add job-scoped `restore-keys`** so a lockfile change still restores the latest matching cache (same job/os/python/poetry); `make dev-deps` reconciles afterward.

## Notes

- Affects the three cache steps that exist today: `unit-tests` `.venv` + pre-commit, and `e2e-tests` `.venv`. The `build` / `release_dev*` jobs have no dependency cache (unchanged), and `release.yaml` has no caches.
- These key changes invalidate existing caches once (one cold run per job), then keys stabilize.
- Independent of open PRs #1030 / #1031 — they don't touch these lines.
- Verified: workflow YAML parses; each rendered key now contains job name + python version (no empty gap) + `poetry<version>`.


minhtule:
> Make keys unique per job via `${{ github.job }}`, so unit-tests and e2e-tests no longer share the same `.venv` key (previously only the first job to finish could save it, and they could cross-contaminate).

Not sure I understand this point. Both `unit-tests` and `e2e-tests` install dependencies using `make dev-deps` and we cache the `.venv` directory so both caches should be the same. Why do we need 2 separate caches? 🤔

minhtule (approved):
LGTM 👍

PR #1033, opened 2026-06-08
fix: mock shard host resolution in unit tests

ekampf:

- Regression from #998

## Changes

- Add an autouse fixture `_mock_shard_resolution` in `app/conftest.py` that patches `app.settings._resolve_shard_host` to a no-op passthrough, so unit tests don't make real network calls when constructing `TwingateOperatorSettings`.
- Patch the resolver inside the module-scoped `_mock_settings` construction as well (module-scoped fixtures set up before the function-scoped autouse, so it can't rely on it).
- Override the fixture with a no-op in `app/tests/test_settings.py` so its dedicated tests keep exercising the **real** `_resolve_shard_host` / `get_host`.

## Notes

For bug fixes:

- **Root cause:** #998 ("feat: Resolve shard URL at Operator startup") added `self.host = get_host(self.network, self.host)` to `TwingateOperatorSettings.__init__` (`app/settings.py:104`). `get_host` → `_resolve_shard_host` does a real `requests.head(...)` retried up to 3× with exponential backoff. `get_host` swallows all errors with a bare `except`, so the calls are silent — but the whole unit suite constructs settings (via the `_mock_settings`/`twingate_settings` fixtures and direct construction), so every run fired many real, retried network requests.
- **Reproduction:** run the unit suite on a machine where `*.twingate.com` HEAD requests are slow/blocked — the suite hangs in retry backoff. With the fix, `make test` runs **269 passed in ~1.2s** (previously minutes, dominated by network/backoff).
- `app/api/tests/conftest.py` and `app/handlers/tests/conftest.py` need no change — the function-scoped autouse covers `twingate_settings`, and the handlers conftest uses a `MagicMock` memo (never constructs settings).

### DIFF
--- a/app/conftest.py
+++ b/app/conftest.py
@@ -13,15 +13,33 @@ def mocked_responses():
         yield rsps
 
 
+@pytest.fixture(autouse=True)
+def _mock_shard_resolution():
+    # Resolving the shard host makes a real (retried) HTTP request on every
+    # TwingateOperatorSettings construction. Make it a no-op in unit tests so
+    # they stay network-free. app/tests/test_settings.py overrides this with a
+    # no-op fixture to exercise the real resolver.
+    with patch(
+        "app.settings._resolve_shard_host", side_effect=lambda network, host: host
+    ):
+        yield
+
+
 @pytest.fixture(scope="module", autouse=True)
 def _mock_settings():
     from app.settings import TwingateOperatorSettings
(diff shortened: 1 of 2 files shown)


minhtule (approved):
LGTM 👍

clement0010 (approved):
LGTM 👍🏻

PR #1034, opened 2026-06-08
chore: upgrade Poetry to 2.4.1 and fix SemVer version handling

ekampf:
The problem gist can be seen here:
```
➜ poetry version 1.2.3-dev.10
Bumping version from 1.1.0 to 1.2.3.dev10
```
poetry now enforces PEP-440 which is different than what we've used so far and it breaks our setup.
This PR updates poetry while keeping old version logic

## Changes

Upgrades Poetry and reworks how the release version is derived so Docker tags and Helm chart versions stay valid SemVer under Poetry 2.4.x's PEP-440 enforcement.

- **Upgrade Poetry 2.2.1 → 2.4.1** (`.tool-versions`, `Dockerfile`) and Python pin in `.tool-versions`.
- **Stop reading the version through `poetry version`.** Poetry 2.4.x normalizes `poetry version` output to PEP-440 (e.g. `1.2.0-dev.123` → `1.2.0.dev123`), which is **invalid SemVer** and rejected by Helm. Now:
  - `Makefile` local `VERSION` default comes from the latest git tag.
  - Dev CI builds the SemVer dev string (`<next>-dev.<run_id>`, dot restored) and pass it via the `VERSION` env — `poetry version` still *writes* pyproject but the value is never read back.
  - Prod release derives the version from the triggering git tag (`${GITHUB_REF_NAME#v}`) for both Docker and Helm.
- **Centralize semantic-release logic into Make targets** so the pinned `python-semantic-release` invocation lives in one place:
  - `make version-next` — next release version (read-only `--print`)
  - `make version-dev BUILD_ID=<id>` — `<next>-dev.<id>`
  - `make version-release` — performs the bump (used by `scripts/release.sh`)
- **Document `.tool-versions`** usage in `CLAUDE.md`.
- **Test fix:** mock shard host resolution in unit tests (`app/conftest.py`, `app/tests/test_settings.py`).

## Notes

- Writing PEP-440 into `pyproject.toml` via `poetry version` remains fine — no code path reads the version at runtime (`package-mode=false`); only Docker/Helm tagging needs SemVer, and those now come from git tags / constructed strings.
- Verified locally: `helm package --version 1.2.0-dev.999` succeeds, while the PEP-440 form `1.2.0.dev999` is rejected with `Error: invalid semantic version` — confirming the fix.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ concurrency:
 
 env:
   PYTHON_VERSION: 3.12
-  POETRY_VERSION: 2.2.1
+  POETRY_VERSION: 2.4.1
 
 jobs:
   dockerfile-lint:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -190,7 +190,7 @@ jobs:
 
       - name: Set DEV release version
         run: |
-          poetry version $(pipx run --spec python-semantic-release==10.5.3 semantic-release --strict --verbose version --print)-dev.${{ github.run_id }}
+          poetry version $(pipx run --spec python-semantic-release==10.5.3 semantic-release --strict --verbose version --print)-dev${{ github.run_id }}
           poetry version -s
 
       - name: Login to Docker Hub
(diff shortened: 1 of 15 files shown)


minhtule (approved):
LGTM 👍

clement0010 (approved):
LGTM 👍🏻

Issue #1037, opened 2026-06-09
Support multiple PAM resources with a Gateway


minhtule:
## Summary

Support multiple PAM resources with a Gateway, including new Web App resource type.

## Details

Introduce new CRDs and a new integration with the Gateway Helm chart:

⚠️ BREAKING CHANGE: remove existing support for Kubernetes resource type: remove `Resource.proxy` field and annotation `resource.twingate.com/type` on k8s services doesn't allow `Kubernetes` type.

**New CRDs**

- `TwingateCertificateAuthority` — registers an X509 certificate authority in Twingate (sourced from a `kubernetes.io/tls` Secret) that signs the Gateway's TLS certificates.
- `TwingateGateway` — represents an in-cluster Gateway endpoint, backed by a Kubernetes Service (the operator resolves its address) and linked to a `TwingateCertificateAuthority`.

**TwingateResource changes**

- Add a `gatewayRef` field so a Twingate resource can be served through a `TwingateGateway`, replacing the deprecated `proxy` field.
- Add new `WebApp` type

**Helm**

- Integrate with the Gateway Helm chart so the Gateway can be deployed and managed alongside the operator.

PR #1038, opened 2026-06-10
feat: Support `TwingateGateway` and `TwingateCertificateAuthority` custom resources

minhtule:

## Changes

Add the `TwingateGateway` and `TwingateCertificateAuthority` CRDs, their handler logic, and tests.

### `TwingateCertificateAuthority`

Syncs a Twingate X509 CA from the `ca.crt` of a referenced `kubernetes.io/tls` Secret.

- **Immutable fields:** `spec.name`, `spec.secretRef` (enforced via CEL `self == oldSelf`). `spec.id` is operator-managed.
- **Reconciles on:**
  - `create` / `resume` and a periodic timer — create the backend CA if missing and store its ID in `spec.id`.
  - **`ca.crt` content change** on the referenced Secret (indexed Secret → CA). The backend has no CA-update mutation, so a new fingerprint **re-creates** the CA (new `spec.id`; name timestamp-suffixed to avoid duplicates). The old backend CA is left for the backend cleanup cron to reap.
  - `delete` — deletes the backend CA; retries (5s) while still "in use" by a Gateway.

### `TwingateGateway`

Creates/updates a Twingate Gateway whose address comes from a Kubernetes Service and whose TLS is signed by a referenced CA.

- **Immutable fields:** none — `serviceRef` and `x509CertificateAuthorityRef` are mutable; `spec.id` is operator-managed.
- **Status:** `address` (resolved `serviceRef` host:port) and `x509CaId` (resolved CA `spec.id`).
- **Reconciles on:**
  - `create` / `resume` / `spec` update and a periodic timer — resolve the Service address + CA ID, then create/update the backend Gateway.
  - **CA `spec.id` change** (indexed CA → Gateway) — propagates a re-created CA's new ID onto dependent Gateways promptly instead of waiting for the timer.
  - `delete` — deletes the backend Gateway; retries (5s) while still in use by a Resource.

### Tests

- Integration test `test_gateway_flows.py` drives CA → Gateway create, cert rotation propagation, and ordered delete against the live backend.
- Unit tests for both handlers.


tests_integration/test_resource_flows.py
```diff
@@ -18,21 +19,6 @@
 )
 
 
-def create_tls_secret(secret_name, base64_ca_cert):
```
minhtule, line 21: Moved to `tests_integration.utils` to reuse

app/api/client_gateways.py
```diff
@@ -0,0 +1,152 @@
...
+class TwingateGatewayAPIs:
+    def get_gateway(self: TwingateClientProtocol, gateway_id: str) -> Gateway | None:
+        try:
+            result = self.execute_gql(
+                GraphQLRequest(QUERY_GET_GATEWAY, variable_values={"id": gateway_id})
+            )
+            return Gateway(**result["gateway"]) if result["gateway"] else None
+        except TransportQueryError:
+            self.logger.exception("Failed to get gateway")
+            return None
```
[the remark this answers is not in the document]
minhtule: This is ok. We're using the same approach as other existing clients

app/handlers/handlers_certificate_authorities.py
```diff
@@ -0,0 +1,83 @@
...
+    if ca_id := spec.get("id"):
+        client = TwingateAPIClient(memo.twingate_settings, logger=logger)
+        try:
+            if not client.x509_certificate_authority_delete(ca_id):
+                # Transport/API error - retry so we don't leak the backend CA.
+                raise kopf.TemporaryError(
+                    "Failed to delete certificate authority, retrying.", delay=30
+                )
+        except GraphQLMutationError as gqlerr:
+            # GC teardown order is not guaranteed: the CA may still be referenced
```
clement0010, line 109: What is `GC teardown`? 🤔

app/handlers/handlers_certificate_authorities.py
```diff
@@ -0,0 +1,83 @@
...
+            f"ca.crt not found yet in Secret "
+            f"'{ca_spec.secret_ref.namespace}/{ca_spec.secret_ref.name}'.",
+            delay=30,
+        )
+
+    ca = client.x509_certificate_authority_create(
+        name=ca_spec.name, certificate=certificate
+    )
+    patch.spec["id"] = ca.id
+    kopf.info(body, reason="Success", message=f"Created on Twingate as {ca.id}")
```
clement0010: nit: `Created Certificate Authority on Twingate ...` would be nicer to differentiate with the logs in gateway handler

app/handlers/handlers_gateways.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+        return success(twingate_id=gw_spec.id)
+
+    gateway = client.gateway_create(
+        address=address,
+        remote_network_id=settings.remote_network_id,
+        x509_ca_id=x509_ca_id,
+    )
+    patch.spec["id"] = gateway.id
+    kopf.info(body, reason="Success", message=f"Created on Twingate as {gateway.id}")
```
clement0010: ditto: `Created Gateway on Twingate....`

app/handlers/handlers_gateways.py
```diff
@@ -0,0 +1,123 @@
...
+@kopf.on.update("twingategateway", field="spec")
+def twingate_gateway_create_update(body, spec, logger, memo, patch, **kwargs):
+    logger.info("twingate_gateway_create_update: %s", spec)
+
+    # If ID changed from `None` to a value we just created it - no need to
+    # immediately update. (Mirrors the group handler.)
+    diff = kwargs.get("diff")
+    # fmt: off
+    if diff and len(diff) == 1 and diff[0][:3] == ("add", ("spec", "id",), None):
+        return success()
```
clement0010: I think this check is not exactly right for the update handler because the diff path is relative to the field filter (`field="spec"`). See [docs](https://docs.kopf.dev/en/stable/walkthrough/diffs/#diffs).

In other words, we should compare `diff[0][:3] == ("add", ("id",), None)` instead

app/handlers/handlers_gateways.py
```diff
@@ -0,0 +1,123 @@
...
+    if gw_spec.id:
+        if client.get_gateway(gw_spec.id) is None:
+            # Backend entity was deleted out from under us - recreate it.
+            gateway = client.gateway_create(
+                address=address,
+                remote_network_id=settings.remote_network_id,
+                x509_ca_id=x509_ca_id,
+            )
+            patch.spec["id"] = gateway.id
+            kopf.info(body, reason="Success", message=f"Recreated as {gateway.id}")
```
clement0010: Technically, `get_gateway(gw_spec.id)`  could fail and return `None`due to API error but the gateway exists in Twingate. Then calling `gateway_create()` will just return the same gateway, not exactly recreating it. So, this message is not exactly accurate. Perhaps we could compare the old ID and the new ID to check whether the Gateway is indeed recreated

app/handlers/handlers_gateways.py
```diff
@@ -0,0 +1,123 @@
...
+                # Transport/API error - retry so we don't leak the backend Gateway.
+                raise kopf.TemporaryError(
+                    "Failed to delete gateway, retrying.", delay=30
+                )
+        except GraphQLMutationError as gqlerr:
+            # Retry while the Gateway is still referenced by a Resource (GC order
+            # is not guaranteed) rather than leaking the backend entity.
+            if "being used" in gqlerr.message:
+                raise kopf.TemporaryError(
+                    "Gateway still in use, retrying.", delay=30
```
clement0010: If I do `helm upgrade` to disable gateway, this might take at least 30 seconds if the `TwingateGateway` is removed before `TwingateResource` right? Is 30 seconds too long? 🤔

app/api/tests/test_client_certificate_authorities.py
```diff
@@ -0,0 +1,167 @@
...
+                    {"variables": {"id": "ca-id"}}, strict_match=False
+                )
+            ],
+        )
+        result = api_client.get_certificate_authority("ca-id")
+        assert result is not None
+        assert result.id == "ca-id"
+        assert result.name == "My CA"
+
+    def test_get_certificate_authority_not_found_returns_none(
```
clement0010: Lets add a test to check `get_certificate_authority()` returns `None` on `TransportQueryError`

app/api/tests/test_client_gateways.py
```diff
@@ -0,0 +1,175 @@
...
+                    {"variables": {"id": "gw-id"}}, strict_match=False
+                )
+            ],
+        )
+        result = api_client.get_gateway("gw-id")
+        assert result is not None
+        assert result.id == "gw-id"
+        assert result.address == ADDRESS
+
+    def test_get_gateway_not_found_returns_none(
```
clement0010, line 33: ditto

app/api/tests/test_client_certificate_authorities.py
```diff
@@ -0,0 +1,167 @@
...
+                    "x509CertificateAuthorityCreate": {
+                        "ok": False,
+                        "error": "some error",
+                    }
+                }
+            }
+        )
+        mocked_responses.post(test_url, status=200, body=failed_response)
+        with pytest.raises(
+            GraphQLMutationError,
```
clement0010, line 146: We should add a test for `TransportQueryError` as well since we handle them differently

app/utils_k8s.py
```diff
@@ -54,3 +55,90 @@ def k8s_read_namespaced_secret(
...
+
+    Raises ``kopf.TemporaryError`` (so the handler retries) when the referenced
+    object does not exist yet or has not finished syncing to the backend.
+    """
+    obj = k8s_get_twingate_custom_object(plural, namespace, name)
+    if obj is None:
+        raise kopf.TemporaryError(
+            f"Referenced {plural} '{name}' in namespace '{namespace}' not found.",
+            delay=delay,
+        )
```
clement0010, line 117: We should probably set a max retries limit for the handler calling this method. Otherwise, if the Twingate CR indeed does not exists due to typo, the handler would retry indefinitely. 🤔

app/handlers/handlers_gateways.py
minhtule: > Technically, `get_gateway(gw_spec.id)` could fail and return `None` due to API error but the gateway exists in Twingate.

I think we only get `TransportQueryError` when the input `ID` is invalid (e.g. malformed base64). If it's a network issue, we should get a different error and the handler should retry. 

> Then calling `gateway_create()` will just return the same gateway, not exactly recreating it. So, this message is not exactly accurate.

When calling the `gateway_create()`, there is always the possibility that the CA already exists even when the `TwingateGateway` CR is created for the first time (i.e. `spec.id = None`). If we want to know if the Gateway is actually created or reused, we need the GraphQL to return that information e.g. adding a `created` boolean field. I think we can keep it simple and add that field later if we really need it (more than just for logging).

app/handlers/handlers_gateways.py
minhtule: I agree. Reduced this to 5s.

app/handlers/handlers_certificate_authorities.py
minhtule: GC = [k8s garbage collection](https://kubernetes.io/docs/concepts/architecture/garbage-collection/)

app/api/tests/test_client_certificate_authorities.py
minhtule: I added `test_x509_certificate_authority_delete_transport_error_returns_false` test for the delete operation. This test is for the create operation which treats all errors the same (it doesn't handle any error).

app/utils_k8s.py
minhtule: Added in https://github.com/Twingate/kubernetes-operator/pull/1038/commits/83fa17267cb78fb395c56a568b63640819aaff0e

app/api/tests/test_client_certificate_authorities.py
minhtule: Added `test_get_certificate_authority_transport_error_returns_none`

app/api/tests/test_client_gateways.py
minhtule: Added `test_get_gateway_transport_error_returns_none`

tests_integration/test_gateway_flows.py
```diff
@@ -0,0 +1,103 @@
...
+            name: {svc_name}
+            port: {port}
+          x509CertificateAuthorityRef:
+            name: {ca_name}
+    """
+
+    with run_kopf() as runner:
+        try:
+            # The CA handler reads ca.crt from this kubernetes.io/tls Secret.
+            kubectl_create(create_tls_secret(secret_name, generate_base64_ca_cert()))
```
minhtule, line 71: Use a newly created CA cert here instead of the hardcoded `BASE64_OF_VALID_CA_CERT` because that cert is used in the [`test_kubernetes_resource_flows`](https://github.com/Twingate/kubernetes-operator/blob/303e6d79718723b4f4cf59cf1dd3c36b2bb20c2b/tests_integration/test_resource_flows.py#L104-L209). The Gateway and CA are not cleaned up in that test so if we reuse the CA here, it can't be deleted, causing the test to fail.

app/handlers/handlers_gateways.py
```diff
@@ -0,0 +1,127 @@
...
+    patch.spec["id"] = gateway.id
+    kopf.info(
+        body, reason="Success", message=f"Created Gateway on Twingate as {gateway.id}"
+    )
+    return success(twingate_id=gateway.id)
+
+
+# Bound retries so a bad ref (e.g. a typo) eventually fails instead of retrying
+# forever; the timer reconciler still recovers it if the ref is later fixed.
+GATEWAY_HANDLER_TIMEOUT = int(os.environ.get("GATEWAY_HANDLER_TIMEOUT", timedelta(minutes=10).seconds))  # fmt: skip
```
clement0010, line 68: Should we add the same timeout for CA handler too?

clement0010 (approved):
LGTM 👍🏻

app/handlers/handlers_certificate_authorities.py
```diff
@@ -0,0 +1,77 @@
...
+    return success(twingate_id=ca.id)
+
+
+@kopf.on.resume("twingatecertificateauthority")
+@kopf.on.create("twingatecertificateauthority")
```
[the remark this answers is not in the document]
ekampf: If we dont support update we need to add validations to CRD to actually block it - otherwise user will make edits and wont know we actually do nothing.

app/handlers/handlers_certificate_authorities.py
```diff
@@ -0,0 +1,83 @@
...
+
+def _reconcile_certificate_authority(body, spec, logger, memo, patch):
+    ca_spec = CertificateAuthoritySpec(**spec)
+
+    client = TwingateAPIClient(memo.twingate_settings, logger=logger)
+
+    # Already registered and still present on the backend - nothing to do.
+    # NOTE: the backend has no CA update mutation, so a rotated cert / renamed CA
+    # is reconciled by re-creating it (a new Twingate ID), not by updating in place.
+    if ca_spec.id and client.get_certificate_authority(ca_spec.id):
```
ekampf: "reconcile" should check server state matches what we think it should be.
It should reconcile server state with what operator knows the correct.

ekampf: if we cant reconcile atleast we should communicate an error (and set the status of the resource to error)

app/handlers/handlers_gateways.py
```diff
@@ -0,0 +1,127 @@
...
+        "twingatecertificateauthorities",
+        gw_spec.x509_certificate_authority_ref.namespace,
+        gw_spec.x509_certificate_authority_ref.name,
+    )
+    patch.status["address"] = address
+
+    if gw_spec.id:
+        if client.get_gateway(gw_spec.id) is None:
+            # Backend entity was deleted out from under us - recreate it.
+            gateway = client.gateway_create(
```
ekampf: its ugly that this code and line 45-50 duplicated.
If we have id and it exists on server we should call `update` and return.
if not it should continue to lines 45-50 which create...

deploy/twingate-operator/crds/twingate.com.twingategateway.yaml
```diff
@@ -0,0 +1,78 @@
...
+                      type: string
+                      description: "Name of the TwingateCertificateAuthority object."
+                    namespace:
+                      type: string
+                      default: default
+                      description: "Namespace of the TwingateCertificateAuthority object."
+            status:
+              type: object
+              x-kubernetes-preserve-unknown-fields: true
+      additionalPrinterColumns:
```
ekampf, line 57: Add CA id

app/handlers/handlers_certificate_authorities.py
```diff
@@ -0,0 +1,171 @@
...
+            _reconcile_certificate_authority(
+                ca_obj, ca_obj["spec"], logger, memo, patch
+            )
+        except Exception:
+            logger.exception(
+                "Failed to reconcile certificate authority %s/%s after secret change",
+                ca_namespace,
+                ca_name,
+            )
+            continue
```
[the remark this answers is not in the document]
minhtule: Event handler doesn't retry ([logs](https://docs.kopf.dev/en/stable/handlers/))

> If the event handler fails, the error is logged to the operator’s log, and then ignored.

app/handlers/handlers_gateways.py
minhtule: Yes, added!

deploy/twingate-operator/crds/twingate.com.twingategateway.yaml
minhtule: Added!

tests_integration/test_gateway_flows.py
```diff
@@ -0,0 +1,144 @@
...
+            kubectl_apply(
+                create_tls_secret(secret_name, generate_base64_ca_cert("Rotated CA"))
+            )
+
+            ca = kubectl_wait_object(
+                "tgca",
+                ca_name,
+                lambda o: o["spec"]["id"] != original_ca_id,
+                description="a new spec.id after ca.crt rotation",
+            )
```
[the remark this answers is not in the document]
clement0010: We should fix this

app/api/client_certificate_authorities.py
```diff
@@ -0,0 +1,116 @@
...
+from pydantic import BaseModel
+
+from app.api.exceptions import GraphQLMutationError
+from app.api.protocol import TwingateClientProtocol
+
+
+class CertificateAuthority(BaseModel):
+    id: str
+    name: str
+    fingerprint: str | None = None
```
clement0010: I think we can probably make `fingerprint` a required field since it's also required in the schema. It would simplify some of the checks in the handler

tests_integration/test_gateway_flows.py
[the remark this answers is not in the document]
minhtule: In this flow that tests a changed value for `ca.crt` in the k8s secret, the `TwingateCertificateAuthority` already has an existing `spec.id` that would be changed. So the `id` key should exist here.

app/handlers/handlers_gateways.py
```diff
@@ -0,0 +1,220 @@
...
+            "Certificate authority %s id changed, reconciling gateway %s/%s.",
+            name,
+            gw_namespace,
+            gw_name,
+        )
+        patch = SimpleNamespace(spec={}, status={})
+        try:
+            _reconcile_gateway(gw_obj, gw_obj["spec"], logger, memo, patch)
+        except kopf.TemporaryError as err:
+            # CA/service not ready yet - retry the event.
```
clement0010:
```suggestion
            # Service not ready yet - retry the event.
```

I think only Service can be in a not-ready state here. The CA should already have an updated twingate ID to invoke this handler. 🤔

app/handlers/tests/test_handlers_certificate_authorities.py
```diff
@@ -0,0 +1,350 @@
...
+    ):
+        # Reconcile blows up (cert not ready yet) - the error is logged and the
+        # patch is not persisted.
+        mock_get_obj.return_value = {
+            "metadata": {"namespace": "default", "name": "my-ca"},
+            "spec": {"name": "My CA", "secretRef": {"name": "gateway-tls"}},
+        }
+        mock_get_certificate.return_value = None
+
+        self._call(self._index())
```
clement0010, line 380: I'd add a second element in `ca_refs` of the index to ensure the handler actually continues through the `ca_ref`

app/handlers/tests/test_handlers_gateways.py
```diff
@@ -0,0 +1,401 @@
...
+        # so the handler does not re-raise and the patch is not persisted.
+        mock_get_obj.return_value = {
+            "metadata": {"namespace": "default", "name": "my-gw"},
+            "spec": {**_spec(with_id=True)},
+        }
+        mock_api_client.get_gateway.side_effect = GraphQLMutationError(
+            "GetGateway", "boom"
+        )
+
+        self._call(self._index())
```
clement0010, line 380: ditto

app/api/client_certificate_authorities.py
```diff
@@ -0,0 +1,116 @@
...
+                GraphQLRequest(
+                    QUERY_GET_CERTIFICATE_AUTHORITY, variable_values={"id": ca_id}
+                )
+            )
+            ca = result["certificateAuthority"]
+            # `certificateAuthority` is a union; the query only selects X509 fields.
+            # A non-X509 type comes back as just `__typename`, so treat it the same
+            # as "not found" rather than letting Pydantic raise on missing fields.
+            if not ca or ca.get("__typename") != "X509CertificateAuthority":
+                return None
```
clement0010, line 80: I think we're missing a test for the `ca.get("__typename") != "X509CertificateAuthority"` condition check

clement0010 (approved):
LGTM! A few minor comments

app/api/client_certificate_authorities.py
minhtule: Updated in https://github.com/Twingate/kubernetes-operator/pull/1038/commits/aeeec83fd3a907aa5ec7263e01752a64e6281c3d

app/handlers/handlers_gateways.py
minhtule: https://github.com/Twingate/kubernetes-operator/pull/1038/commits/960d4e4fff9a4dfeca0b64d74eeafb7ded847c3d

app/handlers/tests/test_handlers_gateways.py
minhtule: Added in https://github.com/Twingate/kubernetes-operator/pull/1038/commits/392dfbfd0fa002add069dc9be638699ef0d1fc44

app/handlers/tests/test_handlers_certificate_authorities.py
minhtule: Added in https://github.com/Twingate/kubernetes-operator/pull/1038/commits/27b1540e7d6989482666c27d704dcb6ef484ec1b

app/api/client_certificate_authorities.py
minhtule: Added in https://github.com/Twingate/kubernetes-operator/pull/1038/commits/5f216c3a6ba5ec95759f337b006b3c8a12a88e31

app/handlers/handlers_certificate_authorities.py
```diff
@@ -0,0 +1,179 @@
...
+    k8s_patch_twingate_custom_object,
+)
+
+
+def _reconcile_certificate_authority(body, spec, logger, memo, patch):
+    ca_spec = CertificateAuthoritySpec(**spec)
+
+    client = TwingateAPIClient(memo.twingate_settings, logger=logger)
+
+    certificate = ca_spec.get_certificate()
```
ekampf: bad name - we should rename to signify this gets it from the `Secret` object in k8s

deploy/twingate-operator/crds/twingate.com.twingatecertificateauthority.yaml
```diff
@@ -0,0 +1,75 @@
...
+                      type: string
+                      description: "Name of the Secret object."
+                    namespace:
+                      type: string
+                      default: default
+                      description: "Namespace of the Secret object."
+            status:
+              type: object
+              x-kubernetes-preserve-unknown-fields: true
+      additionalPrinterColumns:
```
ekampf, line 54: Lets show the secret ref too.
Claude says concatentation could work:
```
- name: Secret
    type: string
    jsonPath: "{.spec.secretRef.namespace}{' '}{.spec.secretRef.name}"
```

deploy/twingate-operator/crds/twingate.com.twingategateway.yaml
ekampf: Lets add serviceRef too see comment on CA CRD

ekampf (approved):
Fix and merge

deploy/twingate-operator/crds/twingate.com.twingategateway.yaml
```diff
@@ -0,0 +1,82 @@
...
```
ekampf, line 57: Lets add secretRef too - see comment above on CA CRD

deploy/twingate-operator/crds/twingate.com.twingatecertificateauthority.yaml
minhtule: While this is a valid JSON path, k8s API server doesn't allow this for column value.
```
The CustomResourceDefinition "twingategateways.twingate.com" is invalid: spec.additionalPrinterColumns[3].JSONPath: Invalid value: "{.spec.serviceRef.namespace}{' '}{.spec.serviceRef.name}": must be a simple json path starting with
```
I'm adding 2 columns: `Secret Namespace` and `Secret`. `Secret Namespace` has [priority](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#priority) 1 so it is only displayed with `-o wide` option (https://github.com/Twingate/kubernetes-operator/pull/1038/commits/7de00cad27347eb83e235745f48058ac73fdad66). For example
```
$ kubectl get tgca                                                                                                                                                           
NAME         ID                                         TYPE   SECRET        AGE
gateway-ca   WDUwOUNlcnRpZmljYXRlQXV0aG9yaXR5OjE4NA==   X509   gateway-tls   93s

$ kubectl get tgca -o wide                                                                                                                                                   
NAME         ID                                         TYPE   SECRET NAMESPACE   SECRET        AGE
gateway-ca   WDUwOUNlcnRpZmljYXRlQXV0aG9yaXR5OjE4NA==   X509   default            gateway-tls   3m11s
```

app/handlers/handlers_certificate_authorities.py
minhtule: Updated to `get_certificate_from_secret()` in https://github.com/Twingate/kubernetes-operator/pull/1038/commits/7beee74075a5bcc8aa495ed1381dccd1de293e9b

deploy/twingate-operator/crds/twingate.com.twingatecertificateauthority.yaml
minhtule: Discussed and decided to set both columns with `priority=1` i.e. both only show up with `-o wide` option

PR #1084, opened 2026-07-15
feat!: Remove deprecated Kubernetes support

minhtule:

## Changes

- Remove Kubernetes-type support from Service annotations
  - `resource.twingate.com/type: Kubernetes` now returns an unsupported-type error
  - Remove `resource.twingate.com/tlsSecret` annotation support
- Remove the deprecated `spec.proxy` field from `TwingateResource`
  - Kubernetes Resources now require `gatewayRef`
- CRD: drop the `proxy` schema; CEL rule reworked to `Kubernetes and WebApp Resources require gatewayRef; Network Resources do not allow it`
- Remove the CA-cert secret-watch handler that only served proxy-based resources

## Notes

- **BREAKING CHANGE**: `spec.proxy` is removed; `type` is immutable so proxy-based Kubernetes Resources cannot be converted in place. 
- Upgrade path: 
  + Disable the Gateway subchart (removes the in-cluster kubernetes-access resource)
    - If `TwingateResource` Kubernetes type is created manually, remove those resources first.
  + `kubectl apply` the new CRDs (Helm does not upgrade `crds/`)
  + Upgrade the operator and re-enable the Gateway subchart


app/crds.py
```diff
@@ -389,10 +315,29 @@ def get_certificate_from_secret(self) -> str | None:
...
+            )
+
+        try:
+            ca_cert = base64.b64decode(ca_cert).decode()
+            validate_pem_x509_certificate(ca_cert)
+            return ca_cert
+        except ValueError as ex:
+            raise kopf.PermanentError(
+                f"Kubernetes Secret object: {secret_name} ca.crt is invalid."
+            ) from ex
```
minhtule, line 361: Duplicated from `ResourceProxy.read_certificate_authority_cert_from_secret()` which is deleted.

app/tests/test_crds_certificate_authority.py
```diff
@@ -83,3 +84,37 @@ def test_get_certificate_returns_none_when_secret_missing(read_secret_mock):
     spec = CertificateAuthoritySpec(name="My CA", secret_ref={"name": "gateway-tls"})
 
     assert spec.get_certificate_from_secret() is None
+
+
+class TestReadCACertFromSecret:
```
minhtule, line 89: `ResourceProxy.read_certificate_authority_cert_from_secret` helper method is moved to `CertificateAuthoritySpec.read_certificate_authority_cert_from_secret` so we need to move the test too.

app/handlers/handlers_secrets.py
```diff
@@ -1,95 +0,0 @@
...
-from app.api import TwingateAPIClient
-from app.api.client_resources import NetworkResource
-from app.api.exceptions import GraphQLMutationError
-from app.crds import ResourceSpec
-from app.handlers.handlers_services import k8s_get_twingate_resource
-from app.utils import validate_pem_x509_certificate
-
-
-@kopf.on.event("", "v1", "secrets", field=("data", "ca.crt"))  # type: ignore[arg-type]
-def twingate_tls_secret_update(
```
minhtule, line 15: This handler is to sync TLS secret changes to `Resource.proxy.certificate_authority_data` so it's no longer needed.

app/handlers/handlers_resource.py
```diff
@@ -154,21 +154,3 @@ def twingate_resource_gateway_index(namespace, name, spec, **_):
             "name": name,
         },
     }
-
-
-@kopf.index("twingateresource")
-def twingate_resource_secret_index(namespace, name, spec, **_):
```
minhtule, line 160: This index is used by `twingate_tls_secret_update` handler which is no longer needed.

app/api/client_resources.py
```diff
@@ -899,24 +844,6 @@ def web_app_resource_update(
...
-            "kubernetesResourceUpdate",
-            GraphQLRequest(
-                MUT_UPDATE_KUBERNETES_RESOURCE,
-                variable_values={
-                    "id": id,
-                    "certificateAuthorityCert": certificate_authority_cert,
-                },
-            ),
-        )
-        return KubernetesResource(**result["entity"])
```
minhtule, line 940: This is used to update the k8s resource's CA certificate when the k8s TLS secret changes. It also uses the deprecated field in `kubernetesResourceUpdate` mutation.

ekampf (approved):

clement0010 (approved):
LGTM 👍🏻

PR #1094, opened 2026-07-22
feat: add remoteNetworkId to TwingateResource and TwingateConnector CRDs

ekampf:

- Issue:

## Changes

- Add `remoteNetworkId` to the **TwingateResource** and **TwingateConnector** CRD spec schemas (string, validated with the same base64 GlobalID `pattern` as `securityPolicyId`)
- Make the field **immutable once set** via an optional-immutable CEL guard: `(!has(oldSelf.remoteNetworkId) || self.remoteNetworkId == oldSelf.remoteNetworkId)` — the same shape the connector CRD already uses for `id`
- Add integration tests (happy path, invalid-pattern rejection, immutability rejection) to `tests_integration/test_crds_connector.py` and `tests_integration/test_crds_resource.py`
- Add unit tests covering the settings-default fallback and the explicit per-object override

## Notes

The Pydantic models already supported this field — `ResourceSpec.remote_network_id` and `ConnectorSpec.remote_network_id` (both defaulting to `get_settings().remote_network_id`, i.e. the operator-wide `TWINGATE_REMOTE_NETWORK_ID`). But because neither CRD schema declared `remoteNetworkId` and the `spec` objects don't set `x-kubernetes-preserve-unknown-fields`, `kubectl` dropped/rejected the field, so it was unreachable in practice. This PR only touches the CRD YAMLs + tests; no Python model changes were needed.

This unlocks running a single operator across **multiple** remote networks by overriding the operator-wide default per Resource/Connector.

Design choices:
- **No CRD-level `default`** — the runtime default must stay the Pydantic default-factory (settings). A YAML default would bake the operator default into stored objects and, combined with immutability, permanently lock them.
- **Not `nullable`** — the Python type is `str` (unlike `securityPolicyId` which is `str | None`), so a literal `null` should be rejected.

### DIFF
--- a/app/tests/test_crds_connector.py
+++ b/app/tests/test_crds_connector.py
@@ -105,6 +105,20 @@ def test_deserialization_imagepolicy(sample_connector_object_imagepolicy):
     assert crd.spec.pod_labels == {"my-label": "my-value"}
 
 
+def test_remote_network_id_defaults_to_settings(sample_connector_object_image):
+    crd = TwingateConnectorCRD(**sample_connector_object_image)
+    # `_mock_settings` (app/conftest.py) sets the operator-wide default.
+    assert crd.spec.remote_network_id == "UmVtb3RlTmV0d29yazoxMjMK"
+
+
+def test_remote_network_id_override(sample_connector_object_image):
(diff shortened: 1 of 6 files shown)


minhtule (approved):
LGTM 👍

Issue #1103, opened 2026-07-27
`TwingateResourceAccess` bindings keep a stale ID for up to 10h after the referenced resource or group is recreated


minhtule:
## Problem

A `TwingateResourceAccess` binding resolves `resourceRef` / `groupRef` to the referenced
object's `spec.id` at reconcile time. Nothing on the binding watches those IDs, so when a
`TwingateResource` or `TwingateGroup` is recreated and gets a new backend ID, the binding
is only re-applied when its 10-hour sync timer next fires (or the binding's own spec is
edited).

Until then:

- access is not granted on the newly created resource/group, so users lose access even
  though the binding objects all look healthy
- the binding's `status` (and its printer columns) keep advertising the old IDs

The same gap applies to a binding created before its referenced object has synced: it
retries on its own, but once the reference does get an ID there is nothing that promptly
re-drives the binding.

## Desired outcome

A binding is re-applied promptly when the resource or group it references gets a new ID,
with the existing sync timer remaining only as a backstop. Long-interval timers should not
be the primary mechanism for propagating an ID change.

PR #1104, opened 2026-07-27
feat: Reconcile `TwingateResourceAccess` when the referenced resource or principal ID changes

minhtule:

- Issue: #1103

## Changes

- Watch `spec.id` on `TwingateResource` and `TwingateGroup`; on change, re-apply the bindings referencing that object instead of waiting up to 10h for the sync timer
- Remove the access granted under the old resource or group ID
- Record the granted pair at `.status.resourceId` / `.status.principalId` instead of inside the create handler's result, which Kopf replaces on every run and so loses the pair whenever a reconcile fails; 
  + The printer columns use these new fields.
- Migrate the recorded IDs to the `status` root on operator start, so bindings last reconciled by an earlier version keep their grant tracked
- Delete using the recorded pair rather than re-resolving `resourceRef` / `groupRef`, so a recreated Group's old grant is revoked and a deleted `TwingateGroup` no longer blocks the binding's own deletion

## Notes

- Only bindings that already carry a recorded grant are re-applied on an ID change; one still waiting for its first grant is left to its own create handler, which retries on its own.


tests_integration/test_resource_flows.py
```diff
@@ -303,3 +305,145 @@ def test_resource_access_flows(
...
+        assert original_resource_id is not None
+
+        kubectl_create(access_obj)
+        access = kubectl_wait_object_handler_success(
+            "tacc", unique_resource_name, "twingate_resource_access_change"
+        )
+        assert (
+            access["status"]["twingate_resource_access_change"]["resource_id"]
+            == original_resource_id
+        )
```
clement0010: nit: nicer to reuse
```suggestion
            _access_status(access).get("resource_id")
            == original_resource_id
```

tests_integration/test_resource_flows.py
```diff
@@ -303,3 +305,145 @@ def test_resource_access_flows(
...
+        assert original_group_id is not None
+
+        kubectl_create(access_obj)
+        access = kubectl_wait_object_handler_success(
+            "tacc", unique_resource_name, "twingate_resource_access_change"
+        )
+        assert (
+            access["status"]["twingate_resource_access_change"]["principal_id"]
+            == original_group_id
+        )
```
clement0010: ditto

clement0010 (approved):
LGTM 👍🏻

ekampf (changes requested):
Not handling deleting the old access.
If I change groups (or resourceid) you need to make sure the old access edge is deleted

minhtule:
Good catch, fixed in a24a78f. `_reconcile_resource_access` now compares the resolved `(resource_id, principal_id)` against the pair recorded in `status.twingate_resource_access_change` and calls `resourceAccessRemove` on the old pair when they differ, after the new grant has been applied. That covers the create/update, timer, and `spec.id` field-handler paths.

Note the CRD marks `resourceRef`, `groupRef`, `principalId` and `principalExternalRef` immutable, so a binding cannot be repointed by editing its own spec; the only way the applied edge changes is the referenced object's `spec.id` changing.

app/handlers/handlers_resource_access.py
```diff
@@ -130,9 +163,18 @@ def twingate_resource_access_sync(
...
+    result = _reconcile_resource_access(body, namespace, spec, status, memo, logger)
+
+    # Kopf files each handler's result under its own handler id, so the timer's result lands
+    # in `twingate_resource_access_sync` while `check_status_created` reads only
+    # `twingate_resource_access_change`. Copy successes across, or the timer never records the
+    # IDs it reconciled and every tick re-issues the same removal. Failures are left out on
+    # purpose: writing `success: False` here would hide the recorded IDs from the next
+    # reconcile, which would then skip the removal and strand the old access.
+    if result.get("success"):
+        patch.status[twingate_resource_access_change.__name__] = result
```
minhtule: I think the right way to do it is to add the principal and resource IDs at the status root. For example,
```
Status:
  principal_id:  U2VydmljZUFjY291bnQ6YmE4MTRiNzEtNThkMC00ZDZkLTlmODItODY2ZWM1NmU0YTRh
  resource_id:   UmVzb3VyY2U6MzQ2MTk5NQ==

  twingate_resource_access_change:
    Success:       true
    Ts:            2026-05-18T15:43:59.319451
  twingate_resource_access_sync:
    Success:       true
    Ts:            2026-08-07T13:34:27.815347
```
instead of this currently

```
Status:
  twingate_resource_access_change:
    principal_id:  U2VydmljZUFjY291bnQ6YmE4MTRiNzEtNThkMC00ZDZkLTlmODItODY2ZWM1NmU0YTRh
    resource_id:   UmVzb3VyY2U6MzQ2MTk5NQ==
    Success:       true
    Ts:            2026-05-18T15:43:59.319451
  twingate_resource_access_sync:
    principal_id:  U2VydmljZUFjY291bnQ6YmE4MTRiNzEtNThkMC00ZDZkLTlmODItODY2ZWM1NmU0YTRh
    resource_id:   UmVzb3VyY2U6MzY1MzY0NA==
    Success:       true
    Ts:            2026-08-07T13:34:27.815347
```

We would need to add some migration logic though.

app/handlers/handlers_resource_access.py
```diff
@@ -88,6 +88,8 @@ def twingate_resource_access_change(
     try:
         client = TwingateAPIClient(memo.twingate_settings, logger=logger)
         principal_id = get_principal_id(access_crd, creation_status, client, namespace)
+        # Delete before adding so that a failure leaves less access than intended.
+        delete_old_access(client, creation_status, resource_id, principal_id, logger)
         client.resource_access_add(
```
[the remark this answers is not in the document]
minhtule: PR description is outdated.

app/handlers/handlers_resource_access.py
```diff
@@ -110,6 +112,37 @@ def twingate_resource_access_change(
         return fail(error=mex.error)
 
 
+def delete_old_access(
```
ekampf, line 128: in previous PR you used `_` prefix to make sure is this kind of method is "private" why not here?

minhtule: No reason really... Even in those PR, there are helper methods that are not "private". The methods used inside those helper methods would be prefixed with `_`. If I touch those methods again, I would remove those `_` prefix to be consistent with existing helper methods (no prefix).

app/handlers/handlers_resource_access.py
ekampf: Both incorrect we shouldnt touch Status or use it for logic.
If we want to track the the content of the references resource there are 2 ways to do it:

1. add "referencedId" to `resourceRef` object and add it when resolving
2. do like kopf and track diff of ref objects via annotations

I like 2 better

minhtule: Discuss offline and we'll store the observed states in the `status` root. The kopf handler's returned values will be kept unchanged.

app/handlers/handlers_resource_access.py
```diff
@@ -110,6 +112,37 @@ def twingate_resource_access_change(
...
+    old_resource_id = creation_status.get("resource_id")
+    old_principal_id = creation_status.get("principal_id")
+    if not old_resource_id or not old_principal_id:
+        return
+
+    if (old_resource_id, old_principal_id) == (resource_id, principal_id):
+        return
+
+    logger.info("Deleting old access %s<>%s", old_resource_id, old_principal_id)
+    client.resource_access_remove(old_resource_id, old_principal_id)
```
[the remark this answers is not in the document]
minhtule: The right fix would be to not swallow network error in the client. We'll fix it separately.

app/handlers/handlers_resource_access.py
```diff
@@ -54,25 +56,26 @@ def get_principal_id(
...
+# before they moved to the root of the status. It is only read as a fallback because Kopf
+# replaces the whole result on the next run, losing the IDs when a reconcile fails.
+LEGACY_ACCESS_STATUS_KEY = "twingate_resource_access_change"
 
-    return None
 
+def get_recorded_access(status: dict | None) -> dict[str, str | None]:
+    """Return the resource and principal IDs recorded for the access grant."""
+    status = status or {}
+    legacy = status.get(LEGACY_ACCESS_STATUS_KEY) or {}
```
ekampf: Can we do `twingate_resource_access_change.__name__` instead of the const?

ekampf (approved):
fix and merge

app/handlers/handlers_resource_access.py
```diff
@@ -130,21 +202,176 @@ def twingate_resource_access_sync(
...
+                ra_obj["spec"],
+                ra_obj.get("status"),
+                memo,
+                logger,
+                patch,
+            )
+
+            # Kopf patches the object the handler is bound to, not the binding.
+            k8s_patch_twingate_custom_object(
+                "twingateresourceaccesses", ra_namespace, ra_name, patch
```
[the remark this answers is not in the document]
minhtule: Concurrent reconciles converge on the pair rather than racing on it. Every path resolves the target from live state (`resource_crd.spec.id`, and the Group's `spec.id` for a `groupRef`), and `resourceRef`, `groupRef`, `principalId` and `principalExternalRef` are all CEL-immutable, so two overlapping runs compute the same target. The second either finds `recorded_id == new_id` and skips, or repeats `remove(old)` and `add(new)`; the removal tolerates `"does not exist"` and re-adding an existing grant is a no-op.

A run holding a stale resource ID cannot overwrite a newer status either. Its `resourceAccessAdd` targets the backend resource that was deleted when the ID changed, `resource_access_add` has no `"does not exist"` tolerance, so it raises and `reconcile_resource_access` returns `fail(...)` before reaching the status patch.

What remains is narrower than an intermediate pair: the mutable spec fields (`expiresAt`, `accessPolicy`, `approvalMode`, `securityPolicyId`) can differ between two overlapping runs, so editing one at the same moment the Resource is recreated can leave the grant on the pre-edit attributes. The pair is still correct, and the next sync re-applies the attributes.

app/handlers/tests/test_handlers_resource_access.py
```diff
@@ -437,95 +492,691 @@ def test_create_passes_approval_mode_to_client(
...
+        kopf_exception_mock.assert_called_once_with(
+            "", reason="Failure", message="resourceAccessRemove failed: some error"
+        )
+
+    def test_transport_error_on_the_deletion_is_retried(
+        self, network_resource_factory, kopf_info_mock, mock_api_client
+    ):
+        # Unlike the GraphQL rejection above, a transient error must propagate instead of
+        # being recorded as a failure, so the handler is retried and the old access is
+        # removed on a later attempt.
```
clement0010: This test looks incomplete, we didn't assert that the old access is indeed removed on a later retry attempt 🤔

clement0010 (approved):
LGTM 👍🏻

app/handlers/tests/test_handlers_resource_access.py
minhtule: This test asserts that the `reconcile_resource_access` should catch `GraphQLMutationError` but not `TransportServerError`. It's kind of redundant and doesn't add any code coverage so I removed it.

PR #1110, opened 2026-07-30
feat: Prepare Gateway service's `TwingateResource` for v2 migration

minhtule:

- Issue: https://github.com/Twingate/kubernetes-operator/issues/1037#issue-4623560178

## Changes

- Set Helm ownership annotation on Gateway service's `TwingateResource` object so Helm can adopt the object when upgrading to v2.
- Add logic to not delete Gateway service's `TwingateResource` when the `resource.twingate.com: true` annotation is removed from the Gateway service

### DIFF
--- a/app/handlers/handlers_services.py
+++ b/app/handlers/handlers_services.py
@@ -41,6 +41,25 @@ UPSTREAM_PORT_ANNOTATION = "resource.twingate.com/upstreamPort"
 REQUEST_HEADER_REWRITES_ANNOTATION = "resource.twingate.com/requestHeaderRewrites"
 
 
+# Metadata that lets Helm take over the generated Kubernetes Resource, which v2's Gateway
+# chart declares under the same name. This operator has to write it while it is still the
+# one running, since Helm checks ownership when the v2 chart is applied.
+# TODO: Remove in v2, which declares and owns the Resource from the start.
+HELM_OWNERSHIP_ANNOTATIONS = (
+    "meta.helm.sh/release-name",
+    "meta.helm.sh/release-namespace",
+)
+HELM_MANAGED_BY_LABEL = "app.kubernetes.io/managed-by"
+# Labels identifying the chart revision that rendered the Service. They describe the
+# Service rather than the Resource derived from it, and copying them makes the operator
+# the owning field manager of values that go stale on the next chart bump, which then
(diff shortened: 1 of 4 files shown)


app/handlers/handlers_services.py
```diff
@@ -309,6 +361,31 @@ def twingate_service_annotation_removed(body, spec, namespace, meta, logger, **_
...
+        # Resource mid-upgrade. Leave it for the v2 chart to adopt. Deleting the Service
+        # outright is unaffected: the owner reference still garbage-collects the object.
+        #
+        # v2 keeps this guard as a safety net. The annotations are expected to be removed
+        # while this operator is still running, but if the removal is only observed after
+        # the upgrade, the event lands on v2, which would then delete the Resource its
+        # chart now owns.
+        # TODO: Remove once no v1 operator is still running in the field, and in v3 at
+        # the latest.
+        if existing_resource_object["spec"].get("type") == ResourceType.KUBERNETES:
```
clement0010, line 375: I think this also applies to hand-crafted Service k8s objects. They'd have to clean up the twingate resource manually after the upgrade? Can we remove it if the twingate resource is not owned by Helm?

clement0010 (approved):
One comment

LGTM 👍🏻

app/handlers/handlers_services.py
minhtule: Note that Helm ownership annotations `meta.helm.sh/release-name`/`-namespace` are written by the Helm client, not by templates, so Argo (helm template + apply) never has them. Keying on them would delete the Resource the v2 chart declares on Argo-managed clusters, which is why the v2 owner-ref release keys on type too.

A hand-crafted Service is possible (a Gateway deployed without the chart), just undocumented. The cost there is small and one-directional: the object is left behind, we log a warning and emit an event saying to delete it explicitly, and deleting the Service still garbage-collects it. Missing a chart-deployed setup would instead deprovision the backend Resource and lose its ID. And the guard goes away once no v1 operator is in the field.

> Can we remove it if the twingate resource is not owned by Helm?

I'm not sure about this question though. If the k8s service is hand-crafted and the annotation is removed, the backing `TwingateResource` would remain and the user has to remove it manually.

ekampf (approved):

minhtule:
We cut a v1.3 on this branch. This can be closed now.

PR #1112, opened 2026-07-30
feat: Repair Kubernetes Resources left behind by the v2 upgrade

minhtule:

## Changes

Repair pre-v2 Kubernetes Resources in place, on both the update handler and the reconcile timer, so an upgraded cluster converges without anyone touching the object.

- Bind a Kubernetes Resource left without a `gatewayRef` to the `TwingateGateway` referencing the Service it was generated from
  - retries while that Gateway does not exist yet
  - refuses a same-named Gateway whose `serviceRef` points at another Service
  - leaves hand-authored names alone (no `-resource` suffix), so those still require an explicit `gatewayRef`
- Drop the Service owner reference from a Kubernetes Resource, leaving the chart as its only owner
  - Network and WebApp Resources the operator still generates from Service annotations keep theirs

## Notes

- Why an upgrade can't fix `gatewayRef` on its own: Helm 3 adopts the existing object once the ownership annotations are set, but never patches custom resources the way it does built-in types (helm/helm#11650), and the field's value is identical in every chart revision, so no later upgrade recovers it. Server-side apply (Helm 4, Argo CD) does write it, so this only repairs what other clients leave behind.
- Why the owner reference has to go: the chart now declares the same object garbage collection owns. Deleting the Service destroys something the release still declares, the client applies it back without a `spec.id`, and the operator registers a second backend Resource.

### DIFF
--- a/app/handlers/handlers_resource.py
+++ b/app/handlers/handlers_resource.py
@@ -43,10 +43,14 @@ def _repair_missing_gateway_ref(name, namespace, spec, patch, logger) -> bool:
 
     # Only adopt a Gateway whose serviceRef points at the Service this Resource was
     # generated from, so a same-named but unrelated Gateway can't capture the access.
-    if gateway.get("spec", {}).get("serviceRef", {}).get("name") != gateway_name:
+    # serviceRef.namespace defaults to the Gateway's own, matching resolve_namespace().
+    service_ref = gateway.get("spec", {}).get("serviceRef", {})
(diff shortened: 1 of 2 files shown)


clement0010 (approved):
LGTM 👍🏻 Maybe we can add integration test to test this flow?

minhtule:
> Maybe we can add integration test to test this flow?

A full integration would tricky. We need the v1 operator logic as well as v1 CRD, besides many mocks and patching.

ekampf (approved):

PR #1122, opened 2026-08-07
fix: Recreate `TwingateResource` when the k8s service's `resource.twingate.com/type` annotation value changes

minhtule:

## Changes

- Delete and re-register the generated `TwingateResource` when a Service's `resource.twingate.com/type` changes; `spec.type` is immutable and no backend mutation converts a Resource's type, so the in-place update was rejected and retried foreverl.

## Notes

- The recreated Resource has a new Twingate ID, so `TwingateResourceAccess` re-grants group access on its next reconcile rather than immediately

### DIFF
--- a/app/handlers/handlers_services.py
+++ b/app/handlers/handlers_services.py
@@ -198,6 +198,10 @@ def twingate_service_create(body, spec, namespace, meta, logger, reason, **_):
     if existing_resource_object := k8s_get_twingate_resource(
         namespace, resource_object_name, kapi
     ):
+        _delete_if_resource_type_changes(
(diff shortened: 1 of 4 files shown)


ekampf (approved):

app/handlers/handlers_services.py
```diff
@@ -231,6 +235,47 @@ def twingate_service_create(body, spec, namespace, meta, logger, reason, **_):
         )
 
 
+def _delete_if_resource_type_changes(
```
ekampf: as dfiscussed remove the `_`

clement0010 (approved):
LGTM 👍🏻

Issue #1128, opened 2026-08-17
Operator >= 1.2.0 crashloops on Amazon EKS: API-server TLS fails with "Missing Authority Key Identifier" (Python 3.13+ `VERIFY_X509_STRICT`)


anirudhbiyani:
## Summary

Starting with operator image `1.2.0`, the Twingate Kubernetes Operator cannot establish a TLS connection to the Kubernetes API server on **any** Amazon EKS cluster. kopf retries the initial `GET /apis` nine times and the pod then crashloops, so no `TwingateResource`, `TwingateResourceAccess`, or `TwingateConnector` custom resources are reconciled.

The cause is a Python-level default change, not an OpenSSL difference: Python 3.13 added `ssl.VERIFY_X509_STRICT` to the default `verify_flags` of `ssl.create_default_context()`. That flag enforces RFC 5280 structurally and requires an Authority Key Identifier (AKID) extension on CA certificates. Amazon EKS cluster CA certificates do not carry an AKID. Operator images from `1.2.0` onward are built on Python 3.14, and kopf builds its SSL context via `ssl.create_default_context()`, so the handshake fails outright.

Last known-good image: `ghcr.io/twingate/kubernetes-operator:1.1.2` (Python 3.12.13).

## Environment

| Item | Value |
| --- | --- |
| Platform | Amazon EKS (reproduced across multiple AWS accounts and regions) |
| Kubernetes server version | `v1.34.9-eks-bca9cf6` (not version-specific — the cluster CA format is the same across EKS versions) |
| API server endpoint (in-cluster) | `https://172.20.0.1:443` (`kubernetes.default.svc`) |
| Broken images | `ghcr.io/twingate/kubernetes-operator:1.2.0`, `1.3.0`, `2.0.0` |
| Working image | `ghcr.io/twingate/kubernetes-operator:1.1.2` |
| CA source | `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` (projected service-account token) |

## Steps to reproduce

1. Install the Twingate Kubernetes Operator on any Amazon EKS cluster using an operator image `>= 1.2.0`.
2. Watch the operator pod:

   ```bash
   kubectl -n twingate logs -f deploy/twingate-operator
   ```

3. Observe the repeated `ClientConnectorCertificateError` below, then `CrashLoopBackOff` once the retries are exhausted.

[...]

PR #1129, opened 2026-08-18
fix: Relax `VERIFY_X509_STRICT` so the operator can reach the API server on EKS

ekampf:

- Issue: #1128 / [OSS-169](https://linear.app/twingate/issue/OSS-169)

## Changes

- Add `app/ssl_compat.py`, which clears `ssl.VERIFY_X509_STRICT` from the SSL contexts the operator builds for the Kubernetes API server.
- Patch **both** API-server clients: kopf/aiohttp (`kopf.ConnectionInfo.as_ssl_context`) and the `kubernetes` client (urllib3's `create_urllib3_context`).
- Apply it from the `@kopf.on.startup()` handler in `main.py`, which runs to completion before kopf's `credentials retriever` task starts, so it lands before the first login.
- Add `TWINGATE_STRICT_X509_VERIFICATION=true` as an opt-out for clusters with an RFC 5280 compliant CA, documented next to `extraEnvVars` in `values.yaml` (no chart template or CRD change needed).
- Add `app/tests/test_ssl_compat.py` (20 cases).

### Root cause

Python 3.13 added `ssl.VERIFY_X509_STRICT` to the default `verify_flags` of `ssl.create_default_context()`, and urllib3 2.x [mirrors that default](https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/ssl_.py). The flag enforces RFC 5280 structurally, including an Authority Key Identifier extension on the CA certificates in the chain.

Amazon EKS cluster CAs are issued **without** an AKID — the reporter verified this across 12+ clusters in multiple AWS accounts and regions; they carry only `Key Usage` and `Basic Constraints`. Operator images `>= 1.2.0` are built on Python 3.14 (`1.1.2` was 3.12.13), so every TLS connection to the API server fails and the pod crashloops:

```
SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed:
Missing Authority Key Identifier (_ssl.c:1081)')
```

This is a Python default, not an OpenSSL difference — the working `1.1.2` image and the broken `2.0.0` image link the identical OpenSSL 3.5.6. Rebuilding against a different OpenSSL would not help.

### Steps to reproduce

Install any operator image `>= 1.2.0` on an EKS cluster and watch `kubectl -n twingate logs -f deploy/twingate-operator`. #1128 also has a standalone two-pod A/B probe that isolates the failure from the operator entirely.

### This does not weaken verification

`CERT_REQUIRED`, `check_hostname`, chain validation and expiry checking all stay in force. Only the RFC 5280 structural strictness is dropped, restoring exactly what every release through `1.1.2` shipped:

```
before kopf   : True
before urllib3: True
after kopf    : False verify_mode= 2 check_hostname= True   # 2 == ssl.CERT_REQUIRED
after urllib3 : False
```

### Why both clients, not just kopf

The issue only shows kopf failing, because the pod crashlooped before any handler ran. But `app/utils_k8s.py` reads pods, secrets, deployments and services through the `kubernetes` client, which goes to the same API server over urllib3 and hits the same strict default. `kubernetes/client/rest.py` passes only `cert_reqs`/`ca_certs` to `PoolManager` and exposes no `ssl_context` hook, so that path needs its own patch — a kopf-only fix would start the operator and then fail on the first reconcile.

Note that `create_urllib3_context` is bound by name in three modules (`urllib3.util.ssl_`, `urllib3.util`, `urllib3.connection`), so all three bindings are rebound; there is a test asserting this specifically. Contexts where the caller passed explicit `verify_flags` are left alone.

### Alternatives considered

- **Patch `ssl.create_default_context` globally / a `sitecustomize.py` shim in the image** — larger blast radius, and invisible from the source tree. The targeted patches keep the change to the two clients we own and are unit-testable.
- **Apply at import time** — would give unit tests a global SSL side effect just for importing handlers. The startup handler runs early enough.
- **Pin the base image back to Python 3.12** — gives up 3.14 and only defers the problem.
- **Fix upstream in kopf** — worth raising, but does not cover the `kubernetes`/urllib3 path and would not unblock users on any timeline they can act on.

### Follow-up: 1.3.x backport still needed

This PR targets `main` (2.x) only. The [v1 → v2 migration guide](https://github.com/Twingate/kubernetes-operator/wiki/Migration-v1-to-v2) routes users through chart `1.3.0` to stamp Helm ownership metadata onto the gateway's `TwingateResource` — and `1.3.0`'s image is *also* Python 3.14, so **EKS users currently cannot take the documented migration path at all**. This commit needs a cherry-pick onto a `1.3.x` patch release to unblock that.

@anirudhbiyani offered to test a candidate image against their EKS fleet — worth taking them up on it before release.

### DIFF
--- a/app/ssl_compat.py
+++ b/app/ssl_compat.py
@@ -1,0 +1,140 @@
+"""Compatibility shim for Python 3.13+ strict X.509 verification.
+
+Python 3.13 added :data:`ssl.VERIFY_X509_STRICT` to the default verification
+flags of :func:`ssl.create_default_context`, and urllib3 2.x mirrors that
+default in ``create_urllib3_context``. The flag enforces RFC 5280 structural
+requirements, one of which is an Authority Key Identifier extension on the CA
+certificates in the chain.
+
+Amazon EKS cluster CAs are issued without an AKID, so every TLS connection to
+the API server fails with
+``SSLCertVerificationError: Missing Authority Key Identifier`` and the operator
+crashloops before it can reconcile anything. Both API server clients the
+operator uses are affected: kopf (aiohttp, via
+:meth:`kopf.ConnectionInfo.as_ssl_context`) and the ``kubernetes`` client
+(urllib3, used by :mod:`app.utils_k8s`).
+
+Clearing the flag restores the behavior of Python <= 3.12, which every operator
+release up to 1.1.2 shipped with. Chain validation, expiry checks,
+``CERT_REQUIRED`` and hostname verification all stay in force - only the RFC
+5280 structural strictness is dropped.
+
+Set ``TWINGATE_STRICT_X509_VERIFICATION=true`` to keep the Python 3.13+ default
+on clusters whose CA is RFC 5280 compliant.
+
+See https://github.com/Twingate/kubernetes-operator/issues/1128
+"""
+
+import functools
+import logging
+import os
+import ssl
+from typing import Any
+
+from app.utils import to_bool
+
+STRICT_ENV_VAR = "TWINGATE_STRICT_X509_VERIFICATION"
+
+_PATCHED_MARKER = "__twingate_x509_strict_relaxed__"
+
+
+def is_strict_x509_verification_enabled() -> bool:
+    """Whether the user opted back into the Python 3.13+ strict defaults."""
+    value = os.environ.get(STRICT_ENV_VAR, "").strip()
+    if not value:
+        return False
+
+    try:
+        return to_bool(value)
+    except ValueError:
+        logging.getLogger(__name__).warning(
+            "Ignoring unparseable %s=%r; assuming false.", STRICT_ENV_VAR, value
+        )
+        return False
+
+
+def relax_x509_strict(context: ssl.SSLContext) -> ssl.SSLContext:
+    """Clear ``VERIFY_X509_STRICT`` on ``context``, leaving all else intact."""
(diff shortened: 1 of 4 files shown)


app/ssl_compat.py
```diff
@@ -0,0 +1,140 @@
...
+        logger.info(
+            "%s is set - keeping strict RFC 5280 certificate verification. "
+            "Note that Amazon EKS cluster CAs carry no Authority Key Identifier "
+            "and will be rejected.",
+            STRICT_ENV_VAR,
+        )
+        return False
+
+    patched = _patch_kopf()
+    patched = _patch_urllib3() or patched
```
ekampf, line 130: @clement0010 is this necessary? try running on RKS without patching urllib3

clement0010: Yes, this is needed because `urllib3` on Python 3.13+ is the culprit causing the issue so we need to patch the to overwrite the `verify_flags`. 

Though I think there's a more elegant solution, like overwriting the Kubernetes Client used by Kopf ([example](https://github.com/kubernetes-client/python/issues/2394#issuecomment-3460128887)). Or use a `sitecustomize.py` to patch `ssl.SSLContext.verify_flags` ([example](https://github.com/kubernetes-client/python/issues/2394#issuecomment-4217562130)).

clement0010:
Closing this PR in favor of https://github.com/Twingate/kubernetes-operator/pull/1134

PR #1134, opened 2026-08-20
fix: Relax `ssl.VERIFY_X509_STRICT` so the operator can reach the API server on EKS

### LINKED ISSUE #1128: Operator >= 1.2.0 crashloops on Amazon EKS: API-server TLS fails with "Missing Authority Key Identifier" (Python 3.13+ `VERIFY_X509_STRICT`)

clement0010:

- Fixes #1128

## Changes

- Add opt-in env var `TWINGATE_DISABLE_API_SERVER_STRICT_X509_VERIFICATION=true` (default unchanged): registers a kopf login handler that wraps the default `login_via_client` and clears only `ssl.VERIFY_X509_STRICT` on the Kubernetes API SSL context.

## Notes
This change is only applicable to Amazon Elastic Kubernetes Service (EKS) clusters created before `v1.17` and are being maintained & upgraded to this day. Newer clusters created after v1.17 shouldn't be affected by this issue.


app/auth.py
```diff
@@ -0,0 +1,58 @@
...
+def login_without_strict_x509(
+    *, logger: logging.Logger | logging.LoggerAdapter, **kwargs
+) -> kopf.ConnectionInfo | None:
+    """Authenticate like kopf's default login, minus ``VERIFY_X509_STRICT``.
+
+    ``kopf.login_via_client`` is the only login handler kopf registers by
+    default here (the official ``kubernetes`` client is installed), so
+    wrapping it keeps the authentication behavior identical.
+    """
+    info = kopf.login_via_client(logger=logger, **kwargs)
```
clement0010, line 49: This is the default login we use when we don't specify the login handler. From the [docs](https://docs.kopf.dev/en/stable/authentication/#piggybacking), there are couple of other options to login but I don't think we have `pykube-ng` installed

deploy/twingate-operator/values.yaml
```diff
@@ -145,6 +145,15 @@ priorityClassName: ""
 #   - name: FOO
 #     value: "bar"
 #
+# On EKS clusters created before Kubernetes 1.17, the cluster CA lacks the
+# RFC 5280 extensions that Python 3.13+ requires by default, so the operator
+# cannot connect to the Kubernetes API server
+# (https://github.com/Twingate/kubernetes-operator/issues/1128). Opt out of
+# the strict check:
```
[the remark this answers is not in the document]
clement0010: We verified it that this only happens for clusters created before 1.17. Newer clusters version with this issue is because it's upgraded from pre-1.17 version

deploy/twingate-operator/values.yaml
```diff
@@ -145,6 +145,15 @@ priorityClassName: ""
...
+# extraEnvVars:
+#   - name: TWINGATE_STRICT_X509_VERIFICATION
+#     value: "false"
```
minhtule: This envvar sounds like strict verification is disabled by default. I would name this `TWINGATE_K8S_API_SERVER_STRICT_X509_VERIFICATION_DISABLED` to make it explicitly that only requests to the k8s API server has strict verification disabled.

app/auth.py
```diff
@@ -0,0 +1,58 @@
...
+
+from app.utils import to_bool
+
+logger = logging.getLogger(__name__)
+
+STRICT_X509_VERIFICATION_ENV = "TWINGATE_STRICT_X509_VERIFICATION"
+
+
+def is_strict_x509_verification_disabled() -> bool:
+    value = os.environ.get(STRICT_X509_VERIFICATION_ENV, "true")
```
minhtule: When the envvar is omitted, it's nicer that the default value is falsy. Also can we use `False` boolean value here?

minhtule (approved):
Fix & merge!

app/tests/test_auth.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_non_strict_connection_info_clears_only_the_strict_flag():
+    baseline = kopf.ConnectionInfo(server="https://localhost").as_ssl_context()
+
+    context = NonStrictX509ConnectionInfo(server="https://localhost").as_ssl_context()
+
+    assert not context.verify_flags & ssl.VERIFY_X509_STRICT
+    # Guard against over-clearing: only the strict bit may go — intermediate-CA
+    # trust, certificate validation, and hostname checks must stay enforced.
+    assert context.verify_flags == baseline.verify_flags & ~ssl.VERIFY_X509_STRICT
```
minhtule: I think the 2nd assertion is stricter than the 1st one so the 1st assertion is redundant?

deploy/twingate-operator/values.yaml
clement0010: Updated the naming in [60f1075](https://github.com/Twingate/kubernetes-operator/pull/1134/commits/60f10759cc63b52ac185c1a382bb16ef577eea97)

app/tests/test_auth.py
clement0010: You're right, the second assertion is stricter, ill drop the first one

deploy/twingate-operator/values.yaml
```diff
@@ -145,6 +145,15 @@ priorityClassName: ""
 #   - name: FOO
 #     value: "bar"
 #
+# On EKS clusters created before Kubernetes 1.17, the cluster CA lacks the
+# RFC 5280 extensions that Python 3.13+ requires by default, so the operator
+# cannot connect to the Kubernetes API server
+# (https://github.com/Twingate/kubernetes-operator/issues/1128). Opt out of
+# the strict check:
+# extraEnvVars:
+#   - name: TWINGATE_K8S_API_SERVER_STRICT_X509_VERIFICATION_DISABLED
```
ekampf: `TWINGATE_DISABLE_API_SERVER_STRICT_X509_VERIFICATION`

ekampf (approved):
rename the env var and merge

PR #1138, opened 2026-08-25
chore: Refuse to run integration tests outside the `minikube` context

ekampf:

- Issue: n/a

## Changes

- Add a session-scoped autouse `_require_safe_kube_context` fixture in `tests_integration/conftest.py` that `pytest.exit()`s unless the current kubectl context is `minikube`. `_load_crds` depends on it so fixture ordering can't regress.
- Add `check-kube-context` as a `test-int` Make prerequisite for a faster, clearer failure.
- Add `get_allowed_kube_contexts()` / `get_current_kube_context()` to `tests_integration/utils.py`, reusing the existing `kubectl()` wrapper so the check honours `KUBECTL_COMMAND`.
- Document the safeguard in `DEVELOPER.md` and `CLAUDE.md`.

## Notes

Integration tests drive a real cluster — `_load_crds` applies the CRDs and `run_kopf` teardown runs `kubectl delete twingate --all` — but nothing checked *which* cluster. A typical laptop kubeconfig has `minikube` in the same list as `twingate-prod-k8s-app-us-central1-prod-s1`, one `use-context` away.

The conftest fixture is the actual safeguard (covers `pytest -m integration` and IDE runs, not just `make test-int`); the Make target is only a nicer error. `TWINGATE_ALLOWED_KUBE_CONTEXTS` (comma-separated) is the escape hatch for deliberately targeting another cluster. Both guards fail closed when there's no current context.

CI is unaffected: `scripts/minikube-smoketests.sh` already runs `minikube start`.

Verified with a throwaway `KUBECONFIG`: both guards block a `fake-prod` context and an empty kubeconfig, the escape hatch lets it through, `test_crds_group.py` still passes against real minikube, and `make test` is untouched.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -63,6 +63,7 @@ pinned versions. Keep this file as the single source of truth — when bumping P
 
 - Unit tests use `mocked_responses` and fixtures from `app/conftest.py`
 - Integration tests need `TWINGATE_API_KEY`, `TWINGATE_NETWORK`, `TWINGATE_REMOTE_NETWORK_ID`
+- Integration tests only run against the `minikube` kubectl context (override with `TWINGATE_ALLOWED_KUBE_CONTEXTS`)
 - Mock settings before importing handlers (loaded at module import time)
 
 ## Code Quality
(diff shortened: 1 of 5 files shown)


clement0010 (approved):
LGTM 👍🏻

minhtule (approved):
LGTM 👍

PR #1162, opened 2026-09-17
feat: Support `configMapRef` on `TwingateCertificateAuthority`

clement0010:

- Issue: https://github.com/Twingate/kubernetes-operator/issues/1161

## Changes

- Add `configMapRef` field to `TwingateCertificateAuthority` to read the CA certificate (`ca.crt`) from a ConfigMap, as an alternative to `secretRef`. Exactly one of the two must be set, and the choice cannot be switched afterwards.
  - The operator watches the referenced ConfigMap and re-registers the CA when its certificate changes, as it already does for Secrets.
- The operator's RBAC now includes ConfigMaps.


examples/certificate-authority.yaml
```diff
@@ -5,9 +5,12 @@ metadata:
 spec:
   name: Example Gateway CA
   type: X509
-  # kubernetes.io/tls Secret whose `ca.crt` is registered with Twingate as the
-  # X509 CA. Provisioned by the gateway chart (self-signed or cert-manager) or
-  # an external tool - the operator only reads the public certificate.
+  # Exactly one of `secretRef` or `configMapRef` names the object whose `ca.crt`
+  # is registered with Twingate as the X509 CA. Both are immutable, and the choice
+  # between them cannot be changed after creation.
   secretRef:
```
minhtule: nit: This reads a bit unnatural to me. I would rephrase it to
```
  # Exactly one of `secretRef` or `configMapRef` must be set. It points to the
  # Secret or ConfigMap whose `ca.crt` key is registered with Twingate as the
  # X509 CA. The reference is immutable: neither its target nor the choice of
  # `secretRef` versus `configMapRef` can change after creation.
```

deploy/twingate-operator/crds/twingate.com.twingatecertificateauthority.yaml
```diff
@@ -48,6 +53,21 @@ spec:
                       type: string
                       nullable: true
                       description: "Namespace of the Secret object. Defaults to the TwingateCertificateAuthority's own namespace when omitted."
+                configMapRef:
+                  type: object
+                  required: ["name"]
+                  description: "Reference to the ConfigMap the CA's public certificate (ca.crt, PEM) is read from. Alternative to secretRef."
```
minhtule: the "(ca.crt, PEM)" part doesn't read naturally because it contains 2 different things (the key name and the format). Something like "..., under the `ca.crt` key in PEM format" would be more helpful.

app/handlers/handlers_certificate_authorities.py
```diff
@@ -214,3 +220,36 @@ def twingate_ca_tls_secret_update(
...
+    )
+
+
+@kopf.on.event("", "v1", "configmaps", field=("data", "ca.crt"))  # type: ignore[arg-type]
+def twingate_ca_config_map_update(
+    event, namespace, name, memo, logger, twingate_ca_reference_index, **_
+):
+    # Every namespace has a kube-root-ca.crt ConfigMap with data.ca.crt, so this
+    # fires for far more objects than CAs reference; the index lookup is what keeps
+    # those a no-op.
```
minhtule: Not sure how helpful this comment is, especially "far more objects than CAs reference". Any `on.event` handler with a `field=` presence filter fires for every matching object, not just the referenced ones; the Secret handler above has the same property and no comment. The only ConfigMap-specific fact is that `kube-root-ca.crt` gives every namespace a match, so I'd either say just that or drop the comment.

app/crds.py
```diff
@@ -342,14 +353,41 @@ class CertificateAuthoritySpec(BaseModel):
     name: str
     type: CertificateAuthorityType = CertificateAuthorityType.X509
     # Secret (kubernetes.io/tls) the CA's public certificate (`ca.crt`) is read from.
-    secret_ref: _KubernetesObjectRef
+    secret_ref: _KubernetesObjectRef | None = None
+    # ConfigMap the CA's public certificate (`ca.crt`, PEM) is read from.
```
minhtule: "under the `ca.crt` key in PEM format" would be more helpful than "(ca.crt, PEM)

minhtule (approved):
LGTM 👍

deploy/twingate-operator/templates/clusterrole.yaml
```diff
@@ -25,7 +25,7 @@ rules:
     verbs: [ "*" ]
 
   - apiGroups: ["*"]
-    resources: [pods, services, secrets, services/status]
+    resources: [pods, services, secrets, configmaps, services/status]
```
minhtule: We only need read permissions (list, get, watch) for configmaps right?

app/handlers/handlers_certificate_authorities.py
clement0010: Dropped the comment

deploy/twingate-operator/templates/clusterrole.yaml
clement0010: Yes, separated the permission here: [87c9be6](https://github.com/Twingate/kubernetes-operator/pull/1162/commits/87c9be6a8c5a37e497802bdea3e348056d30f014)

deploy/twingate-operator/crds/twingate.com.twingatecertificateauthority.yaml
```diff
@@ -48,6 +53,21 @@ spec:
                       type: string
                       nullable: true
                       description: "Namespace of the Secret object. Defaults to the TwingateCertificateAuthority's own namespace when omitted."
+                configMapRef:
+                  type: object
+                  required: ["name"]
+                  description: "Reference to the ConfigMap the CA's public certificate under the ca.crt key in PEM format."
```
ekampf: @copilot can you review this help text?

ekampf (changes requested):
As discussed we dont really need a config-map you can still create a secret manually if needed.
This just adds double logic - more we need to test, document and explain.

clement0010:
Closing this PR in favor of https://github.com/Twingate/gateway/pull/554

